Issue #504, opened 2015-06-10
center of rotation defined by viewpoint is not transformed by parent transform


andreasplesch:
A viewpoint can be translated and rotated by a transform node. While the position and orientation of the viewpoint is correctly transformed, the center of rotation defined by the same viewpoint is not transformed by a parent transform node. Here is an example:

http://andreasplesch.github.io/x3dom/CenterOfRotation/transformedviewpoint.html

The box and the viewpoint is wrapped in a transform node with a translation along the x axis. Rotate the camera in examine mode. The centerOfRotation should be translated along with the position of the viewpoint and the box but it is still at the default position of 0,0,0 which is where the small sphere is for reference.

andreasplesch:
This could be probably fixed by applying the current transform to the cor in the getCenterOfRotation function in viewpoint.js:
return this.getcurrenttransform().multpnt(this._vf.centerofrotation) or similar

However, this is not how the current transform gets applied to the position, so there may be a better place elsewhere for this.

andreasplesch:
The main consumer of getCenterOfRotation appears to be Viewarea.js . There, the function is expected to return the world coordinates of the center of rotation in a number of places. Therefore, it does seem cleanest to apply the current transform in the node code for Viewpoint, Orthoviewpoint, GeoViewpoint and Viewfrustum.

andreasplesch:
Thinking more in terms of the spec. and ahead, the centerOfRotation field is an  [IN, OUT] field and should always contain the local coordinates. Therefore it may be necessary to define a setCenterOfRotationWC(position) where position is expected in world coordinates. Currently, the setCenterOfRotation(position) sets the position in world coordinates. Similary, there should be a getCenterOfRotationWC() function which returns the position in world coordinates while the getCenterOfRotation() returns the position in local coordinates.

andreasplesch:
http://andreasplesch.github.io/x3dom/CenterOfRotation/fixed.html

shows that the proposed patch above works in this situation.

Pull request https://github.com/x3dom/x3dom/pull/505

implements.

andreasplesch:
patch for GeoTransform in pull request #506

PR #505, opened 2015-06-11
transform center of rotation

andreasplesch:
This adds support for parent transform of viewpoint node affecting the location of the center of rotation.


andreasplesch:
addresses issue https://github.com/x3dom/x3dom/issues/504

andreasplesch:
Thanks !
Something to keep in mind is that setCenterOfRotation may need review as well.

andreasplesch:
Hm, this does not work anymore.

PR #506, opened 2015-06-11
apply current transform in getCenterOfRotation()

andreasplesch:
and
sync resetView() with Viewpoint.js


andreasplesch:
see https://github.com/x3dom/x3dom/issues/504

andreasplesch:
Thanks.

PR #513, opened 2015-07-01
GeoTransform implementation

andreasplesch:
Please ignore previous pull request as  this pull request can be automatically merged.

This implements fully the GeoTransform node as specified in the ISO spec.. An additional, experimental option is provided as a field with the name globalGeoOrigin. The option was discussed on the geospatial x3d mailing list.


andreasplesch:
Thanks.

PR #526, opened 2015-07-29
SFImage.copy function bug

andreasplesch:
The .setPixels() expects a pixels array, but the image array was provided.
Instead of converting to and from Pixels, it may be possible to just copy the array using the slice method  ?


andreasplesch:
Thanks.

PR #532, opened 2015-08-04
fix527

Sharian:
First Implementation of USE/DEF for BinaryContainers


mlimper:
Thanks - looks like great work, this will be pretty useful.
@tsturm Do you have any additional comments?

Sharian:
Hi, sorry for not keeping up with this. Has this been fixed in the meantime by one of you?

PR #535, opened 2015-08-10
DIS functionality

mcgredonps:
Initial DIS functionality. 

The Distributed Interactive Simulation (DIS) protocol is used in military virtual worlds and mostly revolves around sending entity state updates with an object's position, velocity, and orientation. This is standardized in IEEE 1278.1. The dis.js file implements this packet protocol. 

The ESPDUTransform et al nodes implement the DIS protocol specified in the X3D standard. 

The NetworkSingleton class implements a websocket based transport mechanism to send and receive DIS packets to a websocket server. There is some violence done to the X3D DIS standard, which assumes that the transport mechanism is traditional sockets. This needs to be better addressed in the standard, probably including an alternative WebRTC transport mechanism.


coderextreme:
I’m looking for documentation on how to set up a websocket server.  Right now, I use socket.io, but that has its own protocol.

coderextreme:
I found DISWebGateway!

PR #539, opened 2015-08-21
fix for "value.toFixed() is not a function"

jamjamg:
I had this annoying "value.toFixed() is not a function" when using x3dom in my project, parseFloat works for me, but I don't know about possible side effects

### DIFF
--- a/src/util/States.js
+++ b/src/util/States.js
@@ -64,7 +64,7 @@ x3dom.States = function (x3dElem) {
      */
     this.toFixed = function (value) {
         var fixed = (value < 1) ? 2 : (value < 10) ? 2 : 2;
-        return value.toFixed(fixed);
+        return parseFloat(value).toFixed(fixed);
     };
 
     /**


andreasplesch:
So value was not a number  when the error occured ? 
A quick search indicates that only https://github.com/x3dom/x3dom/blob/master/src/util/States.js#L107
in the update() function seems to call ...States.prototype.toFixed().
Do you have a thought of what the underlying problem was (perhaps no measurements, empty scene or something) ? Just asking because it may make sense to fix whatever caused this error in the first place.
The error may then be actually useful to keep around.

mlimper:
I agree, the actual problem seems to be that the original value was not a number (but, for instance, a string) - in that case, the logic that determines the value of "fixed" is also not working properly, I suppose.

Could you provide some feedback on where the error occurred (and how your scene data looked like)?

Thanks a lot in advance!

jamjamg:
I will try to debug it more deeply and give an reproducable example. At the moment I can say that x3dom is initialized inside an invisible area (hidden div). After reading your comments, i thing it may be caused by an empty ElevationGrid which is not filled with data at the beginning.

jamjamg:
Ok, I have to admit, that my patch does not address the cause of the problem. As I was trying to make a short example, I've seen, that I've loaded this polyfill for Array "find" as well: https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/find

Now in the update() function it happens, that variable "m" contains "find": <code>infoValue.appendChild(document.createTextNode(this.toFixed(measurements[m])));</code> (at least sometimes).

This lets measurements[m] be of type "function" (the find function...). Im not sure what m stands for and why it occasianally contains "find", but it does not like the mozilla array find polyfill to be loaded at the same time... (probably I'm still using it wrong?)

Ecma Script 6 will have the "find" function... so it may become a problem in the future to use "find" as array index here!?

godenji:
Is this PR going to be merged or has the root cause not yet been found?

Several hundred `a.toFixed is not a function` errors and counting (about 5 per second) while page is loaded in Firefox and Chrome.

Issue occurs when including x3dom in a third party site via `<script>...</script>` in the *`<body>` * of the page. Perfectly fine when including x3dom in a standard page via `<head>`.

jamjamg:
As I found out so far, the problem can be fixed by adding several hasOwnProperty()-calls to certain for-loops iterating over arrays. If the found element of an array is not actual an element, but perhaps a function, it then will produce lots of valueToFixed is not a function errors. I gonna provide another fix for this...

tsturm:
Can someone of you build an small example. I can't reproduce the issue.

jamjamg:
Sry, that it took me so long to come back here: I actually tried to get a JSFiddle to show the problem, but I managed to kill my browser every time I tired it.

So here is another way to reproduce it: go to http://examples.x3dom.org/BVHRefiner/BVHRefiner.html and add any array function using the browsers developer console.

e.g. add this:
<pre>
Array.prototype.test= function() {
    // e.g. polyfill for ES6 find
};
</pre>

... and you gonna see lots of "value.toFixed is not a function" in Chrome or FF.

tsturm:
Hi jamjamg,

thanks for the description how I can reproduce the problem.
Issue should be fixed with my last commit 4b23833.

Regards,
Timo

Issue #541, opened 2015-09-06
text rasterized when zoomed


andreasplesch:
Since x3dom uses a texture with the rasterized text as a way to render the text node, the text appears very rasterized when zoomed in. This is a result of using the default settings for canvas text rendering which are only meant to be viewed as one canvas pixel on one screen pixel. This is often not the case in x3dom.
Here is an example from
http://examples.x3dom.org/example/x3dom_text.html
![image](https://cloud.githubusercontent.com/assets/6171115/9702524/d7fd9e18-542e-11e5-8c52-de953fcc2b71.png)

andreasplesch:
The better way to render text would be to vectorize text to polygons (for example using opentype.js). But a quick way to improve the quality of the font rendering is to use ctx,scale() to produce a scaled-up raster where the canvas font renderer can produce a higher resolution font. Here is the improved result from http://andreasplesch.github.io/x3dom/x3dom_text/ using an oversampling factor of 2:

![image](https://cloud.githubusercontent.com/assets/6171115/9702543/06f6aeac-5430-11e5-9556-dcbced069181.png)

andreasplesch:
Here are the changes to Texture.js

https://github.com/x3dom/x3dom/compare/master...andreasplesch:text_work?diff=split&name=text_work

One could make the oversample variable a field ('quality') of the text node.

andreasplesch:
PR #543 addresses this issue.

andreasplesch:
Here is another comparison.
Original:
http://www.web3d.org/x3d/content/examples/Basic/DistributedInteractiveSimulation/CoordinateAxesX3dDISGeoLoc.xhtml
Improved:
http://andreasplesch.github.io/x3dom/x3dom_text/CoordinateAxesX3dDISGeoLoc.xhtml

PR #543, opened 2015-09-06
improve text quality

andreasplesch:
by oversampling by a factor of two but keeping the textured rectangle the same size. Addresses issue #541 . It would be straightforward to make the oversampling factor a custom field of the text node (default 1 or 2) for backwards compatibility.


mlimper:
Nice, thanks - I would agree to your proposal of making this a custom field of the text node.
Could you add the field before we merge this one?

andreasplesch:
Will do. I would prefer to set the default for this custom field ('quality') to 2. I cannot think of a case where the larger texture would be a problem, and it does look much better. Any reason not to set it to 2 ?
I will also bracket the max. to 10 to avoid extremely large textures. And the min. to 0.5.

andreasplesch:
I added a 'quality' field to the FontStyle node with a default of 2.0 and enforced bracketing to 0.5 and 10. FontStyle seemed more appropriate than Text since it really is a style option. I updated the two examples referenced in issue #541 to use the custom quality factors. Let me know if there is anything else.

mlimper:
That's great, thanks a lot for this contribution.

andreasplesch:
Glad to  contribute.

PR #551, opened 2015-09-14
setFieldValue if existing Bool value == false

andreasplesch:
A small fix to address the case:

1) existing node._vf.someBoolean = 'false'
2) try to use node.setFieldValue('someBoolean', true)
3) current if clause in line 141 evaluates to false -> no update (!)
4) instead properly check for existence of field someBoolean as intended

I can probably put together a simple test case if needed.


mlimper:
thanks for the fix.

andreasplesch:
Thanks, I added a test case to the issue.

Issue #545, opened 2015-09-08
vertical and horizontal text positioning off


andreasplesch:
Currently x3dom seems to position a text node too high:
x3dom rendering of 
http://x3dgraphics.com/examples/X3dForWebAuthors/Chapter02-GeometryPrimitives/GeometryPrimitiveNodes.xhtml
![image](https://cloud.githubusercontent.com/assets/6171115/9735106/57b92ae0-5606-11e5-883d-53b23626b93e.png)
expected rendering:
http://x3dgraphics.com/examples/X3dForWebAuthors/Chapter02-GeometryPrimitives/_viewpoints/GeometryPrimitiveNodes.x3d._VP_Default_viewpoint.png
![image](https://cloud.githubusercontent.com/assets/6171115/9735137/89a55830-5606-11e5-82af-d13020de7248.png)
and also too far left.
This is probably a result of x3dom centering the text box at the local origin. By default the origin should be close to the upper left corner.

andreasplesch:
Here is a better defined example:
x3dom rendering of
http://www.web3d.org/x3d/content/examples/ConformanceNist/Appearance/FontStyle/default.xhtml
![stabilebssa14_fig8](https://cloud.githubusercontent.com/assets/6171115/9744680/4b54edf4-563b-11e5-89d6-ab78108c0fe6.png)
expected rendering:
http://www.web3d.org/x3d/content/examples/ConformanceNist/Appearance/FontStyle/_viewpoints/default.x3d._VP_Test_default_FontStyle_rendering.png
![web3d_screen](http://www.web3d.org/x3d/content/examples/ConformanceNist/Appearance/FontStyle/_viewpoints/default.x3d._VP_Test_default_FontStyle_rendering.png)
The text should be to the right of and under the green lines.

mlimper:
Just figured out where in the code the actual placement of the geometry is happening:
https://github.com/x3dom/x3dom/blob/master/src/Texture.js#L550

The _positions buffer holds the four 3D positions of the corners of a screen-aligned quad (triangulated via the respective indices). It seems to add an offset of 0.4 to the y coordinate, for whatever reason (didn't play around with it).

It should be straightforward to change this. However, I'm not sure whether this makes things easier in case you want to center something around, say, a 3D marker. Clearly, it would make it easier if your intended text is supposed to appear next to / under a known 3D point... well, I'm sure the people who wrote the standard already thought about this.

[...]

PR #555, opened 2015-09-17
improve text fontstyle support

andreasplesch:
The PR addresses issue #545. It adds proper positioning and support for all combinations these fields of the FontStyle node:
- justify: added minor alignment, including for FIRST
- topToBottom: added false (bottom to top)

The horizontal=false and leftToRight=false options are not or only partially supported due to lack of these features in canvas text rendering by web browsers.

The default for justify is modified to MIDDLE MIDDLE for backwards compatibility.

Also, the text node length and maxExtent fields are implemented except that text length will only be shortened and never expanded.


mlimper:
thanks

andreasplesch:
Hopefully, most of the examples which use the text node still look good if they use the dev distribution.

PR #559, opened 2015-09-28
Fix error for setEmissiveColor

mvladimirova:
- fix error in setEmissiveColor from xback to xBack


mvladimirova:
@mlimper @tsturm Please check it out. It is a very small fix but without it the function doesn't work.

tsturm:
Hey Marianna, thx for that fix!

mvladimirova:
You are wellcome :)

Issue #560, opened 2015-10-02
escaped quotes in text node


andreasplesch:
Only the first escaped double quote in a text node string is correctly interpreted. Subsequent escaped double quotes retain the backslash when rendered.
Example:
http://andreasplesch.github.io/x3dom/x3dom_text/

andreasplesch:
PR #561 should fix this.
How to enable escaped single quotes ? &apos; is a work around.

PR #561, opened 2015-10-02
fix for escaped quotes in text node

andreasplesch:
addresses #560.
Test by clicking button in http://andreasplesch.github.io/x3dom/x3dom_text/


mlimper:
Thanks for taking care of this issue!

Issue #379, opened 2014-07-14
IndexedFaceSet with concave geometry misrendered


andreasplesch:
The web3d Sri Campus x3d model includes a building represented by an IndexedFaceSet with concave polygons:

https://googledrive.com/host/0BwIhFzkLaQ9XWmJtcDFVMDNKOUk/SriCampus012_concave_geometry.xhtml

x3dom does not quite render the concave polygons correctly although convex is set to 'false'. The IndexedFaceSet does not have repeated, identical points or indexes. Perhaps there are still degenerate triangles generated during triangulation.

InstantPlayer produces a correct rendering:

https://docs.google.com/presentation/d/1bI5cFqdpiRL2wYVmvCUWbj6cxbdjwc1O1WEH4-FBfac

Andreas

andreasplesch:
fixed with #581

Issue #341, opened 2014-05-06
Extrusion caps not working


chrstein:
For rather complex cross-sections the extrusion caps are not generated correctly.

http://x3dom.org/x3dom/test/functional/extrusion.html shows two problematic cases.

Originally, for extrusions with clockwise oriented cross-section vertices the cap 
polygons were created "outside" of the geometry. This was fixed by inverting the list of vertices depending on the "ccw" flag. 

However, this didn't solve the problem for the second case, were the cap simply does not fill the whole cap surface.

![extrusion](https://cloud.githubusercontent.com/assets/2564424/2888867/ce2b500c-d512-11e3-856d-1a4ae7917009.png)

This seems to be resulting from problems in the algorithms of EarClipping.js.

andreasplesch:
issue #379 has a similar issue with an indexedfaceset.

andreasplesch:
fixed: https://rawgit.com/x3dom/x3dom/master/test/functional/extrusion.html

PR #576, opened 2015-11-09
fix Earclipping

andreasplesch:
This PR improves earclipping by using the shoelace method to determine winding order as the sign of the polygon area, and by upping the (arbitrary?) limit of 15 concave (outwards curving) ears to an arbitrary 150.
The shoelace method is both more robust and faster and the limit of 15 concave ears has lead to misrendering.
The new code is considered preliminary. The PR is functional but meant to provide a place for further considerations.


andreasplesch:
The PR fixes these issues:
#379 , fix applied here http://andreasplesch.github.io/x3dom/earclipping/SriCampus012.xhtml
#341 , http://andreasplesch.github.io/x3dom/earclipping/x3dom_extrusion.html

Additionally, this web3d example
http://x3dgraphics.com/examples/X3dForAdvancedModeling/Buildings/Arch.xhtml
renders now correctly
http://andreasplesch.github.io/x3dom/earclipping/arch.xhtml

The concave regression test
https://github.com/x3dom/x3dom/blob/master/test/regression-suite/test/cases/concave/concave.html
still works
http://andreasplesch.github.io/x3dom/earclipping/concave.html

In order to test other scenes I created a GreaseMonkey (TamperMonkey) user script which will apply the fix:
https://gist.githubusercontent.com/andreasplesch/9c7369583dc9f23d6fe7/raw/x3domEarClippingFixer.user.js
Install GreaseMonkey (https://addons.mozilla.org/en-US/firefox/addon/greasemonkey/?src=ss) and click on user script link above to install.
Disable/Enable GreaseMonkey to test fix.

Here are a number of extrusion cap examples which work with and without the fix:

http://www.web3d.org/x3d/content/examples/Basic/HumanoidAnimation/MaskAndSnorkel.xhtml
http://www.web3d.org/x3d/content/examples/Basic/course/SimpleExtrusionShipHull.xhtml
http://www.web3d.org/x3d/content/examples/Basic/course/ExtrusionHeart.xhtml
http://www.web3d.org/x3d/content/examples/Basic/course/ExtrusionExampleShip.xhtml

I would like to test more scenes with more complex polygons as indexedfacesets or caps of extrusions.

andreasplesch:
Before applying the PR it would good to determine why there was a limit of 15 concave ears. It seems to be fail safe to avoid an infinite while loop (which should never happen, regardless) but why 15 ?
Perhaps the intention was to bail after 15 consecutive concave ears which would seem more reasonable but still arbitrary. In this case, resetting the concave ear counter after cutting of a convex ear would be required.

andreasplesch:
It turns out that the cad component editor at http://examples.x3dom.org/editor/component_editor/ is a good tool to test out more complicated cross-sections for extrusions since it has the nifty cross-section editor. It tends to create a lot of points due to the splines. Upping the concave limit to 1500 makes it work better:
stabilebssa14_fig8
fix via greasemonkey.

There are still problems for spirals like this:
stabilebssa14_fig8
This is due to how "outside" ears are determined.
https://github.com/mapbox/earcut probably has a solution.

mlimper:
> Before applying the PR it would good to determine why there was a limit of 15 concave ears. It seems to be fail safe to avoid an infinite while loop (which should never happen, regardless) but why 15 ?

I also have no idea. There have been recent reports that ear clipping did not work as it should.

@dschie If I remember correctly, you (or your student) had issues with the earclipping code, which lead to a self-written workaround for that particular application? If so, could you check if it works as intended with the new x3dom dev version, which includes this pull request?

mlimper:
@andreasplesch Thanks for this pull request, as well as for the detailed review of the current state!

andreasplesch:
Thanks for merging. If all works I would like to clean up the commented out sections and set the concave counter limit to perhaps a 1000. I also found if the counter is reset after finding a real ear, spirals start to work as well. I made this change to the user script gist for testing as well as in my earclipping branch. Here is an example spiral produced with the component editor:

stabilebssa14_fig8

The cross-section has a lot of points and it took a few seconds (long) to generate but it worked.

So I will update this PR perhaps after getting some more feedback and testing.

I will probably use the vrmath2.net logo to generate random polygons for testing.

I think it would make sense to simply adopt the mapbox/earcut.js library, with an adapter or after morphing it into how x3dom uses the current earclipping (getIndices()). The license is very permissive and just requires including a copyright notice.

andreasplesch:
Here is more complicated spiral which took about 4 minutes (!) to finish and come back:
stabilebssa14_fig8
With almost 30k cross-section points, this is a rare case but probably there needs to a way to abort if the triangulation takes too long.

andreasplesch:
I updated my branch and submitted a new PR #577.

PR #577, opened 2015-11-11
further improved Earclipping

andreasplesch:
This PR streamlines and enhances Earclipping further by
- replacing counter based aborting with time out
- avoiding an inner loop as often as possible

At the same time, some code could be removed reducing the total lines of code.
All tests still work and some more complicated polygons now mesh. The time out is set to 5s per polygon and could be adjusted to a higher (or lower) period.

### DIFF
--- a/src/util/EarClipping.js
+++ b/src/util/EarClipping.js
@@ -4,6 +4,7 @@
  *
  * (C)2009 Fraunhofer IGD, Darmstadt, Germany
  * Dual licensed under the MIT and GPL
+ * (C)2015 improvements by Andreas Plesch
  *
  * Based on code originally provided by
  * Philip Taylor: http://philip.html5.org
@@ -14,7 +15,6 @@ x3dom.EarClipping = {
 	
 	reversePointDirection: function (linklist, plane) {
 			var l, k;
-			var count = 0;
 			var z = 0;
 			var nodei, nodel, nodek;
 			
@@ -24,35 +24,18 @@ x3dom.EarClipping = {
 			
 			for (var i = 0; i < linklist.length; i++) {
 				l = (i + 1) % linklist.length;
-				//k = (i + 2) % linklist.length;
 				
 				nodei = linklist.getNode(i);
 				nodel = linklist.getNode(l);
-				//nodek = linklist.getNode(k); 
 				// use standard shoelace			
 				if(plane == 'YZ') {
 					z  += (nodel.point.y - nodei.point.y) * (nodel.point.z + nodei.point.z);
-					//z = (nodel.point.y - nodei.point.y) * (nodek.point.z - nodel.point.z);
-					//z -= (nodel.point.z - nodei.point.z) * (nodek.point.y - nodel.point.y);
 				} else if(plane == 'XZ') {
 					z  += (nodel.point.z - nodei.point.z) * (nodel.point.x + nodei.point.x);
-					//z = (nodel.point.x - nodei.point.x) * (nodek.point.z - nodel.point.z);
-					//z -= (nodel.point.z - nodei.point.z) * (nodek.point.x - nodel.point.x);
 				} else {
 					z  += (nodel.point.x - nodei.point.x) * (nodel.point.y + nodei.point.y);
-					//z  = (nodel.point.x - nodei.point.x) * (nodek.point.y - nodel.point.y);
-					//z -= (nodel.point.y - nodei.point.y) * (nodek.point.x - nodel.point.x);
 				}
-				/*
-				if (z < 0) {
-					count--;
-				} else {
-					count++;
-				}
-				*/
 			}
-			
-			//if (count < 0) {
 			//if counterclockwise
 			if (z > 0) {
 				linklist.invert();
@@ -60,7 +43,7 @@ x3dom.EarClipping = {
 			}	
 			return false;
 	}, 
-
+	
 	getIndexes: function (linklist) {
 		var node = linklist.first.next;
 		var plane = this.identifyPlane(node.prev.point, node.point, node.next.point);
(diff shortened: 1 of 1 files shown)


andreasplesch:
In order to test scenes I created a GreaseMonkey (TamperMonkey) user script which will apply the improved code:
https://gist.githubusercontent.com/andreasplesch/9c7369583dc9f23d6fe7/raw/x3domEarClippingFixer.user.js

ajcyeh:
I can confirm that this PR is an improvement using the Logo program on VRMath2 Editor at 
https://vrmath2.net/sites/default/files/user/guest/logo/Koch_snowflake_extrusion.logo

Currently VRMath2 Editor is using the latest dev version of x3dom with previous fix. You can access https://vrmath2.net/VRM2 and open the Logo program from public folder (no login).

Current fix survived koch 4 1, but not koch 5 1 See below images.
https://vrmath2.net/sites/default/files/user/guest/image/koch_4_1.png
https://vrmath2.net/sites/default/files/user/guest/image/koch_5_1.png

With this improved fix, you can access from https://vrmath2.net/VRM2/?cm=old
It survived koch 5 1, but not koch 6 1. See below images.
https://vrmath2.net/sites/default/files/user/guest/image/koch_5_1_new.png
https://vrmath2.net/sites/default/files/user/guest/image/koch_6_1.png

Another side question, can extrusion be applied to text so to have 3D text?

andreasplesch:
Thanks Andy for testing this out. You could increase the timeout value from 5000 ms to a higher value and make users aware that complicated polygons may take a long time to show.

All these tests use the new PR:

#379, http://andreasplesch.github.io/x3dom/earclipping/SriCampus012.xhtml
#341 , http://andreasplesch.github.io/x3dom/earclipping/x3dom_extrusion.html

http://andreasplesch.github.io/x3dom/earclipping/arch.xhtml

The concave regression test
http://andreasplesch.github.io/x3dom/earclipping/concave.html
http://andreasplesch.github.io/x3dom/earclipping/extrusion.html

andreasplesch:
Text extrusions are not a part of X3D. But there are js libraries which polygonize fonts which you could use for logo ?

ajcyeh:
Thanks for the fix and suggestion. Another test using Hilbert curve at 
https://vrmath2.net/sites/default/files/user/guest/logo/peano_extrusion.logo

Both #576 and #577 fixes have trouble with extrusion. Perhaps it is because there are points in a straight line of the crossSection?

#576 is very fast at https://vrmath2.net/VRM2, but #577 is very slow at https://vrmath2.net/VRM2/?cm=old for this test.

andreasplesch:
#577 just tries harder to complete the triangulation than #576 before it times out. Even increasing the timeout to one minute does not help. So the polygon hits some pathological situation where after most clipping there is a polygon left which cannot be clipped. It must have a convex corner, so the point in triangle (isEar) function thinks there is a point inside although there is not. Probably a point just touches the triangle.

It may be interesting to see how mapbox/earcut deals. Here is the polygon:
[[0 0] [-1 0] [-1 -1] [0 -1] [0 -2] [0 -3] [-1 -3] [-1 -2] [-2 -2] [-2 -3] [-3 -3] [-3 -2] [-3 -1] [-2 -1] [-2 0] [-3 0] [-4 0] [-4 -1] [-5 -1] [-5 0] [-6 0] [-7 0] [-7 -1] [-6 -1] [-6 -2] [-7 -2] [-7 -3] [-6 -3] [-5 -3] [-5 -2] [-4 -2] [-4 -3] [-4 -4] [-4 -5] [-5 -5] [-5 -4] [-6 -4] [-7 -4] [-7 -5] [-6 -5] [-6 -6] [-7 -6] [-7 -7] [-6 -7] [-5 -7] [-5 -6] [-4 -6] [-4 -7] [-3 -7] [-2 -7] [-2 -6] [-3 -6] [-3 -5] [-3 -4] [-2 -4] [-2 -5] [-1 -5] [-1 -4] [0 -4] [0 -5] [0 -6] [-1 -6] [-1 -7] [0 -7] [1 -7] [1 0] [0 0]]
or flattened:
[0 0 -1 0 -1 -1 0 -1 0 -2 0 -3 -1 -3 -1 -2 -2 -2 -2 -3 -3 -3 -3 -2 -3 -1 -2 -1 -2 0 -3 0 -4 0 -4 -1 -5 -1 -5 0 -6 0 -7 0 -7 -1 -6 -1 -6 -2 -7 -2 -7 -3 -6 -3 -5 -3 -5 -2 -4 -2 -4 -3 -4 -4 -4 -5 -5 -5 -5 -4 -6 -4 -7 -4 -7 -5 -6 -5 -6 -6 -7 -6 -7 -7 -6 -7 -5 -7 -5 -6 -4 -6 -4 -7 -3 -7 -2 -7 -2 -6 -3 -6 -3 -5 -3 -4 -2 -4 -2 -5 -1 -5 -1 -4 0 -4 0 -5 0 -6 -1 -6 -1 -7 0 -7 1 -7 1 0 0 0]

andreasplesch:
Although filtering out collinear segments did  not seem to help, using an angle of 88 for the turns in the Hilbert curve can produce a successful triangulation:

stabilebssa14_fig8

This suggests that there is some numerical issue with these kinds of polygons with lots of concave bays of identical size and right angles, and that this seems to be a special case.

Similarly, randomizing the the segment length a bit helps a lot:

stabilebssa14_fig8

https://vrmath2.net/sites/default/files/user/u342/logo/peano_extr.logo

So, to get more robustness as a next step it would makes sense to make the jump to mapbox/earcut as a library rather than reimplementing it. But this would be a more involved effort.

ajcyeh:
Nice play of angle and random length. :-) I have fun playing with this Hilbert curve. Very interesting to see the fixed size and right angle can produce triangulation error. Perhaps these right angles and whole number sizes confused the identifyPlane then misled the isNotEar or isKonvex. mapbox/earcut algorithm looks very promising. 

PS: I saw good debugging in your Logo by using queue list and pd for the line curve. erall is also a good habit too. :-)

mlimper:
Thanks to the two of you, for implementing and testing this.

andreasplesch:
I took a first stab at using mapbox/earcut for x3dom earclipping purposes and the results are quite promising. For complex polygons likes this:
stabilebssa14_fig8
with about 50000 points earcut is very fast, almost instant. Also, there is no timeout required. It may be a little slower for very simple polygons compared to the current earcliipping.
Here is my first implementation as user script:
https://gist.github.com/andreasplesch/42d6405d0de8cc2472b8
It uses mapbox/earcut almost as is, encapsulated into a function in an object. It could live in a separate file.

Would there be interest in using this ? The only requirement is a copyright notice judging from the license language.

ajcyeh:
Thank you Andreas.

I can confirm this earcut is accurate and fast. It can be tested from https://vrmath2.net/VRM2/?cm=old and run this logo program https://vrmath2.net/sites/default/files/user/guest/logo/peano_extrusion.logo

Using this Earcut library there is no need to random size or change 90 degree angle for the Hilbert curve.
peano5

Hope this earcut can be included in x3dom.

andreasplesch:
Here is a level 6 curve with 25000 cross-section (polygon) points which meshes almost instantaneously with earcut.

stabilebssa14_fig8

The recursive Logo generation takes a while but the resulting x3d displays very quickly after it loads.

Issue #277, opened 2014-01-20
missing (incomplete) support for displacement shader in CommonSurfaceShader node


andreasplesch:
The CommonSurfaceShader node currently supports a
diffuse texture map and a normal texture map. See this example

https://googledrive.com/host/0BwIhFzkLaQ9XYlQ0dkRnZm5ZVkk/index.html

However, adding a displacement map results in inconsistent rendering:

https://googledrive.com/host/0BwIhFzkLaQ9XYlQ0dkRnZm5ZVkk/displacement.html

The displacement gets only partially applied (no displacements in Asia
and Europe), a seam opens and the headlight stops working.

Should there be an option for displacement along the
normal ?

andreasplesch:
The fix in #589 is applied to the example above. It works now.

PR #589, opened 2015-12-15
CSS displacement along normal

andreasplesch:
I believe the if clause in the commit is supposed to mean: if there is a displacement map of any kind but no normal map then calculate normals.
The added parentheses are necessary for this meaning. Otherwise
true || false && !true (without parentheses) results in true in javascript (but should be false since there is a normal map).
This fixes essentially #277. (already applied to example).


andreasplesch:
Thanks !

PR #605, opened 2016-01-23
GeoLod implementation

andreasplesch:
This is a (simplistic) implementation for the GeoLOD node. It works with all tested examples on web3d.org which use this node. Once loaded, there is no attempt to unload levels if they are out of range. So beware of very large LODs. Another todo item is to preload levels which are getting close to smooth the transition between levels.


andreasplesch:
Thanks !

PR #606, opened 2016-01-25
FloatVertexAttribute is parsed following indices for indexed nodes

YuanxiangFranck:
Allow the uses of FloatVertexAttribut in geometry like IndexedFaceSet with a more natural input:
The parsing of the FloatVertexAttribut, follows the attribute "index" or "coordIndex"

### DIFF
--- a/src/nodes/Rendering/X3DComposedGeometryNode.js
+++ b/src/nodes/Rendering/X3DComposedGeometryNode.js
@@ -12,7 +12,7 @@ x3dom.registerNodeType(
     "X3DComposedGeometryNode",
     "Rendering",
     defineClass(x3dom.nodeTypes.X3DGeometryNode,
-        
+
         /**
          * Constructor for X3DComposedGeometryNode
          * @constructs x3dom.nodeTypes.X3DComposedGeometryNode
@@ -112,7 +112,7 @@ x3dom.registerNodeType(
              * @instance
              */
             this.addField_SFNode('texCoord', x3dom.nodeTypes.X3DTextureCoordinateNode);
-        
+
         },
         {
             handleAttribs: function()
@@ -120,7 +120,11 @@ x3dom.registerNodeType(
                 //var time0 = new Date().getTime();
 
                 // TODO; handle case that more than 2^16-1 attributes are to be referenced
-                var i, n = this._cf.attrib.nodes.length;
+                var i, j, k, n = this._cf.attrib.nodes.length;
+                var val, index, numComponents, nb_index;
+                // Test if the geometry has indices as index or coordIndex
+                var indices = this._vf.index;;
+                if (!indices) { indices = this._vf.coordIndex; }
 
                 for (i=0; i<n; i++)
                 {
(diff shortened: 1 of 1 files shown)


andreasplesch:
Hm, I think the spec. defines that index or coordIndex should _not_ be used FloatVertexAttributes:

http://www.web3d.org/documents/specifications/19775-1/V3.3/Part01/components/shaders.html#Pervertexattributes

"Per vertex attributes are mapped 1:1 to each vertex value. When used in a node derived from X3DComposedGeometryNode, the number of values defined in the node containing the attribute values shall be identical to the number of coordinate values specified for the geometry node. If the number of values does not match, the visual result is implementation specific."

I think the fix would be to treat FloatVertexAttributes the same as normals, eg. also duplicate them for the creaseAngle=0 case.

YuanxiangFranck:
My mistake I misread the specs, I'm closing the pull request and issue

YuanxiangFranck:
I just tried an more complex example and  creaseAngle=3.14 did not work. ( I don't understand why )

If the FloatVertexAttributes should be treated the same as normals, then you need to parse it following the index. The parsing I wrote is exactly the same as the parsing of the normals in FieldChanged.

I spent more time to study the problem and I think the spec are not clear enough:
> Per vertex attributes are mapped 1:1 to each vertex value. [...]
the number of values defined in the node containing the attribute values shall be identical to the number of coordinate values specified for the geometry node

With my parsing I set 1 data for 1 point defined in the Coordinate node. 
It is mapped 1:1 with the data written in the node, but not with the data computed in the node.

Also without fix, it is impossible to use the FloatVertexAttribute if you don't understand exactly how the IndexedFaceSet are triangulated. 
Exemple : 
```html
<scene>
  <shape id="shape">
    <appearance>
      <Material></Material>
      <ComposedShader id='ComposedShader'>
        <field name='threshold' type='SFFloat' value='0.'></field>
        <!-- Simple Shader to send color and use the data -->
        <ShaderPart type='VERTEX' style="display:none;" >
          #ifdef GL_FRAGMENT_PRECISION_HIGH
          precision highp float;
          #else
          precision mediump float;
          #endif
          attribute vec3 position;
          attribute vec3 color;
          attribute float data;
          uniform mat4 modelViewProjectionMatrix;
          varying float fragData;
          varying vec3 fragColor;
          void main()
          {
          fragData = data;
          fragColor = color;
          gl_Position = modelViewProjectionMatrix * vec4(position,1.);
          }
        </ShaderPart>
        <ShaderPart type='FRAGMENT' style="display:none;">
          #ifdef GL_FRAGMENT_PRECISION_HIGH
          precision highp float;
          #else
          precision mediump float;
          #endif
          uniform float threshold;
          varying vec3 fragColor;
          varying float fragData;

          void main()
          {
          if ( threshold > fragData) { discard; }
          gl_FragColor = vec4(fragColor, 1.0);
          }
        </ShaderPart>
      </ComposedShader>
    </Appearance>
    <IndexedFaceSet id="faceSet" solid="false"
                    coordindex='4 0 1 2 3 -1 0 3 5' >
      <Coordinate point="1.5    0 0,
                         0 -1.5 0,
                         -1.5    0 0,
                         0  1.5 0,
                         0    0 1,
                         2    2 0">
      </Coordinate>
      <Color color="1 0 0, 1 1 0, 0 1 0, 1 1 0, 1 0 0, 1 1 0,  0 1 0,"></Color>
      <!-- creaseAngle='3.4' don't remove duplicated vertices -->
      <FloatVertexAttribute
         name="data" numComponents="1" creaseAngle='3.4'
         value="1 2 3 4 5 6 7 8 ">
        <!--
            Values needed without my parsing :
            value="5 1 2, 2 1 3, 3 1 4, 1 4 6">
          -->
      </FloatVertexAttribute>
    </IndexedFaceSet>
  </shape>
</scene>
```

andreasplesch:
Apologies, I should not have mentioned that FloatVertexAttribute should be treated the same as normals. That was too abbreviated. This is true only in the sense that they need to be duplicated in the creaseAngle=0 case, so that each duplicated vertex has the attribute as it was assigned.

You used creaseAngle as a field for FloatVertexAttribute. There is no such thing. creaseAngle only applies to geometries, eg. IndexedFaceSet.

YuanxiangFranck:
my mistake, wrong copy/past. It worked.
In Indexed faced Set in the attributes are parsed in two ways with a  big if : 
```javascript 
if (((this._vf.creaseAngle <= x3dom.fields.Eps) [...]
```
Should I change my fix this way or should I forget about it?

andreasplesch:
Sorry, what do you have in mind ? But, yes, this may be the right place in IndexedFaceSet to work on.

YuanxiangFranck:
The attributes from FloatVertexAttribute are not treated in each geometry (unlike any others attributes like colors).
It istreated using a call of "handleAttrib" from X3domComposedGeometryNode.
My idea was to change the handleAttrib function, to whether leave it raw (case with ceaseAngle) are trie a parsing (other cases)

PR #610, opened 2016-01-27
setAttribute works on FloatVertexAttribute

YuanxiangFranck:
Fix to allow the users to update the values in FloatVertexAttribute with setAttribute.
I changed the following geometries :  
IndexedFaceSet, IndexedQuadSet, IndexedTriangleSet, IndexedTriangleStripSet, QuadSet, TriangleSet,

Note : For the IndexedFaceSet I did some code cleaning (eebe316),
In the triangulation of the faces the case 2 and case 3 was almost identical, I fused the two cases.

### DIFF
--- a/src/nodes/Geometry3D/IndexedFaceSet.js
+++ b/src/nodes/Geometry3D/IndexedFaceSet.js
@@ -633,7 +633,7 @@ x3dom.registerNodeType(
             {
                 if (fieldName != "coord" && fieldName != "normal" &&
                     fieldName != "texCoord" && fieldName != "color" &&
-                    fieldName != "coordIndex")
+                    fieldName != "coordIndex" && !fieldName.startsWith("attrib_"))
                 {
                     x3dom.debug.logWarning("IndexedFaceSet: fieldChanged for " +
                         fieldName + " not yet implemented!");
@@ -791,8 +791,21 @@ x3dom.registerNodeType(
                         hasColor = false;
                     }
                     this._mesh._numColComponents = numColComponents;
+                    if (fieldName.startsWith("attrib")) {
+                        var attrName = fieldName.slice(7);
+                        var attrNode = this._cf.attrib.nodes.find(
+                            function (a) {return a._vf.name == attrName;}
+                        );
+                        if (attrNode) {
+                            var attr_data = attrNode._vf.value;
+                            var attr_numComp = attrNode._vf.numComponents;
+                            var hasAttributes = this._mesh._dynamicFields[attrName] !== undefined;
+                            this._mesh._dynamicFields[attrName].numComponents = attr_numComp;
+                            this._mesh._dynamicFields[attrName].value = [];
+                        }
+                    }
 
-                    var i, j, t, cnt, faceCnt;
+                    var i, j, k, t, cnt, faceCnt;
                     var p0, p1, p2, n0, n1, n2, t0, t1, t2, c0, c1, c2;
 
                     if(this._vf.convex) {
@@ -930,6 +943,16 @@ x3dom.registerNodeType(
                                             this._mesh._texCoords[0].push(texCoords[t2].z);
                                         }
                                     }
+                                    if (hasAttributes) {
+                                        var pInd = [p0,p1,p2];
+                                        for (j=0;j<3;j++) {
+                                            for (k = 0; k < attr_numComp; k++){
+                                                this._mesh._dynamicFields[attrName].value.push(
+                                                    attr_data[attr_numComp*pInd[j]+k]
+                                                );
+                                            }
+                                        }
+                                    }
                                     p1 = p2;
                                     t1 = t2;
 
@@ -1042,6 +1065,7 @@ x3dom.registerNodeType(
 
                     Array.forEach(this._parentNodes, function (node) {
                         node.setGeoDirty();
+                        if (hasAttributes) node._dirty.specialAttribs = true;
                     });
                 }
                 else {
--- a/src/nodes/Geometry3D/IndexedFaceSet.js
+++ b/src/nodes/Geometry3D/IndexedFaceSet.js
@@ -865,8 +865,6 @@ x3dom.registerNodeType(
                                     else if (hasColorInd && !colPerVert) { c2 = +colorInd[faceCnt]; }
                                     else if (colPerVert) { c2 = p2; }
                                     else { c2 = faceCnt; }
-                                    t = 3;
-
                                     //this._mesh._indices[0].push(cnt++, cnt++, cnt++);
 
                                     this._mesh._positions[0].push(positions[p0].x);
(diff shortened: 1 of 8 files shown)


src/nodes/Geometry3D/IndexedFaceSet.js
```diff
@@ -791,8 +791,21 @@ x3dom.registerNodeType(
                         hasColor = false;
                     }
                     this._mesh._numColComponents = numColComponents;
+                    if (fieldName.startsWith("attrib")) {
+                        var attrName = fieldName.slice(7);
```
andreasplesch: I see. But the name of the attribute is not provided after an underscore ("attrib_temperature") but only in the name field of the attrib child node.

src/nodes/Geometry3D/IndexedFaceSet.js
```diff
@@ -791,8 +791,21 @@ x3dom.registerNodeType(
...
+                        var attrNode = this._cf.attrib.nodes.find(
+                            function (a) {return a._vf.name == attrName;}
+                        );
```
andreasplesch: To find out which attribute changed, it is probably necessary to implement fieldChanged() in FloatVertexAttribute.js .
An option may be to just assume all attribs have changed and update all.

src/nodes/Geometry3D/IndexedFaceSet.js
```diff
@@ -633,7 +633,7 @@ x3dom.registerNodeType(
             {
                 if (fieldName != "coord" && fieldName != "normal" &&
                     fieldName != "texCoord" && fieldName != "color" &&
-                    fieldName != "coordIndex")
+                    fieldName != "coordIndex" && !fieldName.startsWith("attrib_"))
```
andreasplesch: Why use startsWith instead of != and why the underscore in "attrib_" ? I do not think there is a field with such a name ?

src/nodes/Geometry3D/IndexedFaceSet.js
```diff
@@ -865,8 +878,6 @@ x3dom.registerNodeType(
                                     else if (hasColorInd && !colPerVert) { c2 = +colorInd[faceCnt]; }
                                     else if (colPerVert) { c2 = p2; }
                                     else { c2 = faceCnt; }
-                                    t = 3;
-
```
andreasplesch: Not sure. This is setting up for the next index.

src/nodes/Geometry3D/IndexedFaceSet.js
```diff
@@ -932,111 +943,18 @@ x3dom.registerNodeType(
                                             this._mesh._texCoords[0].push(texCoords[t2].z);
                                         }
                                     }
-
```
andreasplesch: not sure. The switch logic is harder to follow, so I am not confident this can be removed. Note the p1=p2;t1=t2 setup. Since there is a lot of code repetition, it would make sense to reuse code in functions. Why not just leave it alone ?

YuanxiangFranck: Before the new indices (p0,p1,p2) where computed before the push of the values, so that why you need to separate the case : 
creation of the first triangle, t=2 -> update p2  -> push -> t=3
creation of the other triangle (t=3): update p0,p1,p2  -> push -> t=3 or t=-1

With the fix : I update p2, then push, then update p0 and p1
Now : update t = 2 -> p2  -> push -> update p0,p1 -> t=2 or t=-1

src/nodes/Geometry3D/IndexedFaceSet.js
YuanxiangFranck: With the fix, there are no case t=3 anymore, the creation of any triangle is set in case t=2

src/nodes/Geometry3D/IndexedFaceSet.js
YuanxiangFranck: The name is given during the call of FiledChanged on the FloatVertexAttribute element
then the FloatVertexAttribute call the fieldChanged from IndexedFaceSet (the code above) with attrib_*name*
see : 16fefd1
```javascript
fieldChanged: function (fieldName) {
                 var that = this;
                 if (fieldName == "value" ||  fieldName == "numComponents") {
                     Array.forEach(this._parentNodes, function (node) {
                         node.fieldChanged("attrib_"+that._vf.name);
                     });
                 }
             }
```

andreasplesch:
This assumes that you have attrib fields with names like "attrib_temperate" in containerfield ?

src/nodes/Geometry3D/IndexedFaceSet.js
YuanxiangFranck: There are no way to know which attribute changed in the FieldChanged of IndexedFaceSet, and this is made specialy to avoid to update/compute all the other attributes.

YuanxiangFranck:
Let's assume the following element : 
<FloatVertexAttribute id="temp" name="temperate" numCompenents="1" value="...."><FloatVertexAttribute>

when temp.setAttribute("value", "....") is called : 
1- FieldChanged (FloatVertexAttribute) is called with "value" as filedName
then it call the FieldChanged (parent node : IndexedFaceSet for exemple)
2- FieldChanged (IndexedFaceSet) is called with "attrib_temperate" as FieldName
That is why I test FieldName.statsWith(""attrib_), then get the name using FieldName.split(7)

This allow to avoid useless computation time by updating all the attributes.

src/nodes/Geometry3D/IndexedFaceSet.js
andreasplesch: I see. While that probably works. I do not think the node.fieldChanged() function is intended to be reused for internal purposes like this. Why not make a new function updateAttrib(name) or in IndexedFaceSet.js (and all possible parentNodes of FloatVertexAttribute) ? Or do the update right here ?

src/nodes/Geometry3D/IndexedFaceSet.js
andreasplesch: I do not have the stomach to think this through. So I would err on the side of caution, and separate this out for now. Somebody else should review this as well. Since this affects most IFSs and most scenes have those, a good test would be see if all web3d and x3dom examples with IFSs still work.

YuanxiangFranck: I edited the explanation to make it easier.
I agree with you, it is safer to test this before merging it.

src/nodes/Geometry3D/IndexedFaceSet.js
YuanxiangFranck: The idea of the new function updateAttrib(name) is not bad, it can be similar to the handleAttibs in X3domComposedGeometryNode.
The only problem is in IFS, I cannot imagine a (proper) way to know which attribute needs update. 
The ugly solution is to save in the node the name of the modified attribute (in _vf for exemple).

andreasplesch: Well, the name in node.updateAttrib(name) would have the name of the changed attribute, no ?

YuanxiangFranck: if you call node.updateAttrib in the IFS, the only information given is FieldChanged("attribs").
A solution would be to write the function node.updateAttrib(name) in FloatVertexAttribute.js, and call it before (or after) the call of FieldChanged on IndexedFaceSet.

PR #615, opened 2016-02-01
Feature inside RadarVolumeStyle

aarbelaiz:
Allowing to set the camera inside the volume when using the RadarVolumeStyle.


andreasplesch:
Thanks. This works and I created a user script which can be included in the html header
 
<script type="text/javascript" src="http://gist.githubusercontent.com/andreasplesch/82fb409cd1338bbf9be9/raw/ec13f552904be8dc8823ee0c6473f2c454c1e45c/RadarVolumeStyleAllowInside.user.js"></script>

or used as Greasemonkey script.

PR #621, opened 2016-02-13
fix text for VR (rendered texture)

andreasplesch:
An empty array for mesh vertex positions leads to a 0 volume for getVolume(). This in turn leads to small feature culling in renderedTexture (and when otherwise requested as it seems to be switched on now by default). This is why text does not display in VR.

The fix simply provides some non-null geometry which leads to initial non-null volume. SInce then text is not culled, the mesh positions get updated to the real geometry (in Texture.js) and getVolume() will get the real volume.


andreasplesch:
Thanks.

PR #627, opened 2016-03-24
Apply devicePixelRatio to canvas coordinates on scaled displays

flatline:
Picking was broken on scaled displays, e.g. Windows scaled to 125% for high DPI display devices. This can easily be seen in the picking demo, but applied equally to any combined use of runtime.mousePosition, runtime.pickRect, or accessing Event.hitPnt.  The coordinates returned for a MouseEvent did not account for the previous use of scaling on the 3d canvas element, resulting in a visible offset between the physical mouse position and the logical position computed by mousePosition.  Scaling the result of this function corrects the issue.


ssdmike:
@flatline @mlimper The result is still not accurate due to the requirement of the css styles and padding. If the css is not set or ignored the pick will not work. 

Avoiding duplicitous code please check out: https://github.com/x3dom/x3dom/issues/675 as I have added the event positions and target offsets to make the x y calculation. this is working on macs and windows with high density displays.

PR #647, opened 2016-06-08
Extendable Navigation Modes and improved Turntable Navigation

skluge:
Most of the Navigation functions from the Viewarea Prototype has been extracted into an new DefaultNavigation Prototype, which can be inherited to create new Navigation Modes.

In addition the current Turntable Navigation has been improved and refactored into a new TurntableNavigation prototype, which inherits from DefaultNavigation.


mlimper:
Thanks a lot!

Panning is still important. It would be nice to have something as in the previous (early) version of the turntable, where you could translate the view plus center of rotation by panning.
When no target is set via scripting, we could then pan without limitation.

andreasplesch:
This looks great. I want to look into using the DefaultNavigation node to define VR navigation modes like in the classroom example. Would this be the right approach ?

skluge:
Sure, i think it would be possible to use the VR device as Input for a new navigation mode.

You should look into the DefaultNavigation prototype and the TurntableNavigation prototype as an example for a new navigation mode.
You also have to register your new navigation mode in the NavigationInfo Node, so it can be used.

skluge:
@mlimper 
As requested it is now possible to use panning without a set bounding box. Now it is also possible to specify the panning axis by code.

andreasplesch:
I looked at the pr more closely but could only make some conceptual progress since a webvr API vrdevice is quite different from a mouse or touches.  For example, navigation in x3dom occurs when mouse/touch/key events are fired. However, webvr does not provide similar events for pose changes. It is probably necessary to hook into the render loop . There is runtime.enterFrame() but this may not be appropriate here. Investigating.

mlimper:
> As requested it is now possible to use panning without a set bounding box. Now it is also possible to specify the panning axis by code.

That sounds really interesting. As you might already know, our current application is this 3D viewer for the berlin museums, which is already online in a first version: http://3d.smb.museum/pergamonaltar/
I added you to the project so you can have a look into the code - I guess it will be beneficial to discuss the features directly by example.

The first thing I would then like to ask at this point is:
How do I set the panning axes programatically? Could you give an example, and maybe even try inside the smb-viewer app (in "interaction3D.js", line 260, there is a function that does this for the "typeParams" of the node - could probably done very similarly, I guess?).

Also, for the small friezes that go around corners (ger.: "Risaliten"), it would be cool to disable panning completely. Is this also possible?

Thanks a lot in advance!

src/nodes/Navigation/modes/TurntableNavigation.js
```diff
@@ -68,40 +71,48 @@ x3dom.TurntableNavigation.prototype.onDrag = function(view, x, y, buttonState)
...
+        cor = view._at;     
                     
-    view._at = cor;
-    
-    // update camera matrix with lookAt() and invert
-    view._flyMat = x3dom.fields.SFMatrix4f.lookAt(view._from, cor, view._up);
-    viewpoint.setViewAbsolute(view._flyMat.inverse());
+        cor = cor.addScaled(this.panAxisX, ty);
+        var temp = cor;
+        if(cor.y > bbox.max.y || cor.y < bbox.min.y )
```
mlimper: I would expect that the x delta on the screen maps to a translation along "panAxisX", not along "panAxisY"

src/nodes/Navigation/modes/TurntableNavigation.js
```diff
@@ -68,40 +71,48 @@ x3dom.TurntableNavigation.prototype.onDrag = function(view, x, y, buttonState)
...
+            temp = view._at;
+        else
+            view._from = view._from.addScaled(this.panAxisX, ty);
+        
+        
+        cor = temp.addScaled(this.panAxisY, tx); 
+        if(cor.x > bbox.max.x || cor.x < bbox.min.x )
+            cor = temp;
+        else
+            view._from = view._from.addScaled(this.panAxisY, tx);
```
mlimper: see comment above

mlimper:
I would expect that the x delta on the screen maps to a translation along "panAxisX" - but it seems to map to one along "panAxisY". Is this intended? Could we change this?

PR #665, opened 2016-09-15
Expose canvas event handlers

DimitarIlievCenterMine:
This PR introduces `X3DCanvas.bindEventListeners()` method and moves all event bindings. 
Before the event handlers are bound they are set as properties to the X3DCanvas instance. The old approach used anonymous functions which prevented developers to extend the default interaction. 

With this PR we are now able to add mouse+keyboard combinations such as __mouseWheel + Alt key__ . Simple implementation could be:

``` js
var d = function(selector) { return document.querySelector(selector)}

x3dom.runtime.ready = function() {
	var x3d = d('x3d');
	var viewpoint = d('viewpoint');
	var x3dCanvas = x3d.runtime.canvas;
	
	var _onMouseWheel = x3dCanvas.onMouseWheel;
	
	x3dCanvas.onMouseWheel = function(event) {
		if(event.altKey) {
			var offset = .01
			var fov = parseFloat(viewpoint.getAttribute('fieldofview')) || 1.571;
			
			if(event.wheelDelta < 0) offset = -offset;
			fov = Math.min(2, Math.max(0, fov+offset));

			viewpoint.setAttribute('fieldofview', fov)

		} else {
			_onMouseWheel.call(this, event);
		}
	}

	x3dCanvas.canvas.removeEventListener('mousewheel', _onMouseWheel); // unable to unbind when handler is anonymous function 
	x3dCanvas.canvas.addEventListener('mousewheel', x3dCanvas.onMouseWheel);
}
```
Another case would be handling rectangular selection and preventing navigation while doing so.

### DIFF
--- a/src/X3DCanvas.js
+++ b/src/X3DCanvas.js
@@ -190,6 +190,177 @@ x3dom.X3DCanvas = function(x3dElem, canvasIdx)
 	this.disableMiddleDrag = x3dElem.getAttribute("disableMiddleDrag");
 	this.disableMiddleDrag = this.disableMiddleDrag ? (this.disableMiddleDrag.toLowerCase() == "true") : false;
 
+    this.bindEventListeners();
+};
+
+
+x3dom.X3DCanvas.prototype.bindEventListeners = function() {
+    var that = this;
+
+    this.onMouseDown = function (evt) {
+        if(!this.isMulti) {
+            this.focus();
+            this.classList.add('x3dom-canvas-mousedown');
+
+            switch(evt.button) {
+                case 0:  this.mouse_button = 1; break;  //left
+                case 1:  this.mouse_button = 4; break;  //middle
+                case 2:  this.mouse_button = 2; break;  //right
+                default: this.mouse_button = 0; break;
+            }
+
+            if (evt.shiftKey) { this.mouse_button = 1; }
+            if (evt.ctrlKey)  { this.mouse_button = 4; }
+            if (evt.altKey)   { this.mouse_button = 2; }
+
+            var pos = this.parent.mousePosition(evt);
+            this.mouse_drag_x = pos.x;
+            this.mouse_drag_y = pos.y;
+
+            this.mouse_dragging = true;
+
+            this.parent.doc.onMousePress(that.gl, this.mouse_drag_x, this.mouse_drag_y, this.mouse_button);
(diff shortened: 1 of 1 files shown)


mlimper:
Thanks for the PR!

I have two questions:

- We need to document this somewhere. There are several places where we document functions or typical workflows, on the one hand the [scene author API docs](http://doc.x3dom.org/author/index.html) and on the other hand the [tutorials](http://doc.x3dom.org/tutorials/index.html). What do you think would be the right place? For the scene author API, there is already a part "Runtime" - theoretically, we could add a similar part "Canvas"... but that seems like a lot for this few functions. For the available mouse events on 3D objects, we have mentioned them in a [specific tutorial](http://doc.x3dom.org/tutorials/basics/htmlCSS/index.html), which works - but could be complemented by an additional reference (under "Scene Author API"). Maybe we could make a general reference for available events?

- A question about your example: For what are the last two lines of code used? It looks like the variables `x3dCanvas.onMouseWheel` and `_onMouseWheel` would point to the same object, so  removing the listener and adding it again seems like a useless operation... is there something I missed here?

mlimper:
Edit, about the second one:
I might have just missed the point in the JS code: you create a pointer to save the old function, then assign a new one, and the first pointer (`_onMouseWheel`) is still pointing to the old function object... is that correct?

DimitarIlievCenterMine:
:+1: This is correct. This way I can add some custom logic when handling the events without breaking the old behaviour. Removing the previous event listener is necessary because the event would always trigger the function that was initial passed as handler.

I looked where would be most appropriate to document it. 
* [Developer API documentation - X3DCanvas](http://doc.x3dom.org/developer/x3dom/X3DCanvas.html) would be perfect place for documenting the `bindEventListeners` and the event handlers
* An [Animations and Interaction Tutorials](http://doc.x3dom.org/tutorials/index.html) would be most suitable with reference to the X3DCanvas Dev documentation.

mlimper:
Thanks a lot. I tried your code and it seems to work nicely. I can't get the example with the viewpoint's FOV working, but that's another issue - at least I can print stuff to the console inside the callback, so your code that adds the custom event handler works.

mlimper:
I added a ticket for adding documentation / tutorial content:
https://github.com/x3dom/x3dom/issues/670

PR #710, opened 2017-01-15
Fix reset of request queues in RequestManager

david-zacharias:
The `abortAllRequests`-function in RequestManager uses one line variable assignment which leads to `this.requests` and `this.activeRequests` referencing the same object after _reset_. Hence in next run requests _active_ as soon as stored on `this.requests`.


mlimper:
Thanks for the fix!

Issue #707, opened 2017-01-10
Texture is not displayed correctly There is a glTF model


cx20:
I tried to display the glTF sample using X3DOM 1.7.3-dev

https://cx20.github.io/gltf-test/examples/x3dom/index.html?category=sampleModels&model=BoxTextured&scale=1&type=glTF-Binary

However, the texture is attached to only one side of the box.
![image](https://cloud.githubusercontent.com/assets/1833979/21813709/7a8d9c10-d79a-11e6-91c2-cde9db781d52.png)

The texture should be affixed to all sides of the cube.
https://github.com/cx20/gltf-test/tree/master/sampleModels/BoxTextured

<img src="https://raw.githubusercontent.com/cx20/gltf-test/master/sampleModels/BoxTextured/screenshot/screenshot.png">

cx20:
The following models also have no texture.

https://cx20.github.io/gltf-test/examples/x3dom/index.html?category=sampleModels&model=Avocado&scale=0.5&type=glTF-Binary
![image](https://cloud.githubusercontent.com/assets/1833979/21854545/736bae9e-d85e-11e6-94ff-680c95aa92cd.png)

https://cx20.github.io/gltf-test/examples/x3dom/index.html?category=sampleModels&model=BarramundiFish&scale=0.05&type=glTF-Binary
![image](https://cloud.githubusercontent.com/assets/1833979/21854618/c0eb2bfe-d85e-11e6-8fca-4e82714bb690.png)

https://cx20.github.io/gltf-test/examples/x3dom/index.html?category=sampleModels&model=SmilingFace&scale=1&type=glTF-Binary
![image](https://cloud.githubusercontent.com/assets/1833979/21854690/0504c340-d85f-11e6-9d22-0df9f3c4ac58.png)

skluge:
Thanks. I will look into this issue.

skluge:
This pull request should solve the texture issues. 
I tested all the mentioned cases and they now seem to work fine.

The hardest one was the cube, because everything seemed to work fine. It was a problem with non power of two textures and in webgl not supported sampling options. These textures are now automatically resized.

@mlimper Please merge pull request

mlimper:
> @mlimper Please merge pull request

Done - thanks for the fixes.

@cx20 Could you try again with the latest dev?

cx20:
I confirmed that it was fixed in the latest version of x3dom.js.
The sample below has been updated.
https://cx20.github.io/gltf-test/

[...]

PR #712, opened 2017-01-17
glTF texture fix

skluge:
Fixed the support for using multiple textures and added support for non-power-of-two textures with resizing when needed.
Should fix all the problems in #707 

Also added a texture cache, so that the glTF Textures are not loaded multiple times.

### DIFF
--- a/src/util/glTF/glTFContainer.js
+++ b/src/util/glTF/glTFContainer.js
@@ -105,6 +105,7 @@ x3dom.glTF.glTFMesh.prototype.render = function(gl, polyMode)
         gl.drawElements(polyMode, this.drawCount, this.buffers[glTF_BUFFER_IDX.INDEX].type, this.buffers[glTF_BUFFER_IDX.INDEX].offset);
     else
         gl.drawArrays(polyMode, 0, this.drawCount);
+
 };
 
 x3dom.glTF.glTFTexture = function(gl, format, internalFormat, sampler, target, type, image)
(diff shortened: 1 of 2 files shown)


src/util/glTF/glTFContainer.js
```diff
@@ -127,43 +128,91 @@ x3dom.glTF.glTFTexture.prototype.isPowerOfTwo = function(x)
...
+x3dom.glTF.glTFTexture.prototype.needsMipMaps = function(gl)
+{
+    var need = true;
+    need &= (this.sampler.magFilter == gl.LINEAR || this.sampler.magFilter == gl.NEAREST);
+    need &= (this.sampler.minFilter == gl.LINEAR || this.sampler.minFilter == gl.NEAREST);
+
+    return !need;
+};
+
+x3dom.glTF.glTFTexture.prototype.nextPowerOfTwo = function(n)
```
tsturm: We have already a utils function to get the next power of two, maybe it make more sense to use that.
https://github.com/x3dom/x3dom/blob/master/src/util/Utils.js#L675

src/util/glTF/glTFContainer.js
```diff
@@ -127,43 +128,91 @@ x3dom.glTF.glTFTexture.prototype.isPowerOfTwo = function(x)
...
+        var width2 = this.nextPowerOfTwo(width);
+        var height2 = this.nextPowerOfTwo(height);
+
+        var aspect2 = width2 / height2;
+
+        if(aspect - aspect2 > 0.01){
+            console.warn("Image "+this.image.src+" needs to be resized to power of two, but has unsupported aspect ratio and may be distorted!");
+        }
+
+        var oc = document.createElement('canvas'),
```
tsturm: At this point you can also use the Utils.scaleImage function
https://github.com/x3dom/x3dom/blob/master/src/util/Utils.js#L634

skluge:
changed the functions as you suggested

Issue #690, opened 2016-12-12
pickValue inconsistent in large scenes and various camera angles


ssdmike:
[macosx][safari][chrome][retina]

I have a large scene >10k m, and when i try to use the shootRay or doubleClick for centering the scene the "picked" position is dramatically incorrect. 

The position error is exaggerated by angle of the camera, looking down creates a very large error as well as looking up.

Also the position of the cursor on the canvas creates exaggerated error as well, the further away from the center of the canvas the error seems to get worse.

I'm trying different things to figure this out but i am out of my element with regards to gl and pixel data. I'm hacking my way through this to make something useable, and so far i have been able to identify the factors of the error:

1. Scene size. Larger scenes = Larger error
2. Camera angle, Looking down and up increases error
3. Position from center of x3d dom element

I am able to visualize the rays and positions and through that i was able to see the errors. I have smoke tested this with small and large and super large scenes >69km.

I am hoping this can be a jumping off point for someone who knows the math and gl better than I do.

i left my references to my visual debug functions makeSphere and makeLine.

```
        // render to texture for reading pixel values
        this.renderPickingPass(gl, scene, mat_view, mat_scene, from, sceneSize, pickMode, x, y, 2, 2);

        // the pixel values under mouse cursor
        var pixelData = scene._webgl.fboPick.pixelData;

        if (pixelData && pixelData.length)
        {
            var pickPos = new x3dom.fields.SFVec3f(0, 0, 0);
            var pickNorm = new x3dom.fields.SFVec3f(0, 0, 1);

            var index = 0;
            var objId = pixelData[index + 3], shapeId;

            var pixelOffset = 1.0 / scene._webgl.pickScale;
            var denom = .003633718561;//1.0 / 256.0; //MR DID THIS BASED ON A LARGE SCENE THAT IS DEFINED BY PLACING BOXES AT KNOWN MIN AND MAX POSITIONS
            var dist, line, lineoff, right, up;

            if (pickMode == 0) {
                objId += 256 * pixelData[index + 2];

                dist = (pixelData[index] / 255.0) * denom + (pixelData[index + 1] / 255.0);
                var dist1 = dist;
                
                line = viewarea.calcViewRay(x, y, cctowc);

                pickPos = line.pos.add(line.dir.multiply(dist * sceneSize));
                

                //makeSphere(pickPos, .3, 'red');
                
                //console.log('pickPos: ' + pickPos);

                index = 4;      // get right pixel
[...]

PR #717, opened 2017-01-29
use zfar for depth scaling in picking

andreasplesch:
if smaller (closer) than total scene size (default). This allows for increasing accuracy of picking by specifying a smaller but reasonable zfar (say 2000) in large scenes (100000 diagonal).
Addresses #690 
This PR only impacts scenes which have a zfar value specified.


andreasplesch:
https://rawgit.com/andreasplesch/x3dom/f5084d5a5e749627972b1c7eb18ec5be676d0146/picking/example.html

has a large scene diagonal and picking is inaccurate.

https://rawgit.com/andreasplesch/x3dom/90e33d574276aeb0098bd028dc420ee2092122a6/picking/example.html

uses the PR and has improved picking accuracy as result.

Issue #709, opened 2017-01-11
glTF model CesiumMilkTruck wheels does not display properly


cx20:
I tried to display the glTF sample using X3DOM 1.7.3-dev. However, the wheels were not displayed correctly.
https://cx20.github.io/gltf-test/examples/x3dom/index.html?category=sampleModels&model=CesiumMilkTruck&scale=0.5&type=glTF-Binary
![image](https://cloud.githubusercontent.com/assets/1833979/21854857/ae469dac-d85f-11e6-8605-df2510e74b1f.png)

I confirmed that it can be displayed by individually settings the mesh of the car body and the wheel like the sample below. However, I think that this model should be displayed without specifying individual settings.

https://examples.x3dom.org/gltf/

skluge:
The x3dom glTF-Loader is only supposed to load mesh and materials.
So the node and transformation tree from the glTF is not used.

cx20:
I understood that the glTF Loader in x3dom does not use nodes and transformation trees yet.

Is it the same reason why Buggy.gltf are not displayed correctly?
https://cx20.github.io/gltf-test/examples/x3dom/index.html?category=sampleModels&model=Buggy&scale=0.02&type=glTF-Binary
![image](https://cloud.githubusercontent.com/assets/1833979/21988466/ac16a66e-dc4a-11e6-8769-f6389d89c8bd.png)

I expect to be displayed as follows.
![image](https://cloud.githubusercontent.com/assets/1833979/21988492/bdfa80bc-dc4a-11e6-9bb3-7e3a00c34180.png)

skluge:
Yes, the problem is, that the meshes are initially positioned in the world center and then transformed to the correct position.
As for the milk truck example, the mesh of the wheels is positioned at the origin, and in the node-tree it is referenced two times at different positions. But since the node tree is not loaded, the mesh stays at the origin.

andreasplesch:
Would the idea of an ExternalGroup help, in addition to ExternalShape, and ExternalGeometry ? Presumably, as a grouping node ExternalGroup could contain a node tree with the Transforms ?

mlimper:
There was already a proposal of making Inline nodes support gltf - could that be a solution?

andreasplesch:
Yes, I think gltf Inlines would be a good solution. Perhaps with namespaced names for the nodes/shapes/geometries so they are available from the main scene ?

andreasplesch:
[...]

PR #723, opened 2017-02-25
use glTF node transform tree

andreasplesch:
This PR constructs a world transform for each glTF node and applies it to the node's meshes and their primitives. The world transforms are factored in during updateTransforms where the appropriate shader uniforms are multiplied with the world transform matrixes.
Some uniforms still need to be addressed. 
Discussed in #709


andreasplesch:
In order to avoid converting transforms for the various uniforms from and to GL format and also avoid a few matrix inversion, it would be more optimal to factor in the node world transforms in gfx_webgl.renderShape, rather than in update Transforms.

The shader parameters there could be recalculated for each of the shape.meshes where these are iterated.

The world transform then could just be kept in a mesh.worldtransform property, rather than material property.

On the other hand gfx_webgl is already long.

andreasplesch:
Ok, I found a mathematical way to avoid additional matrix inversions, eg. keep the existing inversions.

(A x B) ^ -1 = B^-1 x A^-1 (http://mathworld.wolfram.com/MatrixInverse.html)

We want to multiply in the node world transform (B) and can do that by keeping the existing inverse A^-1 and right multiply the inverse of B  which could be precomputed for non-animated nodes.

tsturm:
Thanks Andreas, for this great PR!

andreasplesch:
thanks for merging !
Still missing is to update the picking renderer to also call updateTransforms for glTFs.
Also it would be good to have a getVolume() function for externalShape (similar to externalGeometry).
Are there plans for glTF2 or PBR materials ?

andreasplesch:
missed a few lines while syncing, see #725

PR #737, opened 2017-04-21
Walk mode fix

andreasplesch:
by reinstituting 1.7.1 onDrag behaviour in a separate walk navigation implementation.
Also added back functional walkTrough walk mode test.
https://rawgit.com/andreasplesch/x3dom/walk-mode-fix/test/functional/walkThrough.html


andreasplesch:
Thanks. Here is a slightly modified walk mode with dampened adjustments to ground level which produces a softer, nicer walk:

https://rawgit.com/andreasplesch/x3dom/smooth-walk/test/functional/walkThrough.html
https://github.com/andreasplesch/x3dom/blob/smooth-walk/src/nodes/Navigation/modes/WalkNavigation.js#L387

But I am not sure how the dampening should be controlled (just constant halving seems to work well).

tsturm:
I think the easiest way is to make the dampening configurable over the NavigationInfo.
We could also try to ignore really small distance changes between frames.

tsturm:
I added your proposed walk dampening and make it configurable over the NavigationInfo.
I also fix the strange up-vector of the walk mode.

andreasplesch:
Cool, do you want me to test (somewhere) ?

PR #742, opened 2017-05-10
shininess factor fix

andreasplesch:
use shininessFactor in CommonSurfaceShader as factor for multiplication with texture map value, as described in the node documentation.


jamesleesaunders:
Is this still valid? 
Quickly checked it will not conflict with webVR branch.
https://github.com/x3dom/x3dom/blob/daf3d895d683183f77c5fc2c9499dac4db5e6fcb/src/shader/ShaderDynamic.js#L1056
So perhaps a quick win PR to merge and knock this issue off the list?

andreasplesch:
hm, I think merging this broke automatic merging of webvr into master

Issue #411, opened 2014-09-02
Default texture mapping of the IndexedFaceSet wrongly implemented


fabienrohrer:
When an IndexedFaceSet has no texture mapping (texCoord and texCoordIndex not defined), the texture mapping should be implemented as described there:

http://www.web3d.org/documents/specifications/19775-1/V3.3/Part01/components/geometry3D.html#f-IndexedFaceSettextureDefaultMapping

> If the texCoord field is NULL, a default texture coordinate mapping is calculated using the local coordinate system bounding box of the shape. The longest dimension of the bounding box defines the S coordinates, and the next longest defines the T coordinates. If two or all three dimensions of the bounding box are equal, ties shall be broken by choosing the X, Y, or Z dimension in that order of preference. The value of the S coordinate ranges from 0 to 1, from one end of the bounding box to the other. The T coordinate ranges between 0 and the ratio of the second greatest dimension of the bounding box to the greatest dimension. Figure 13.6 illustrates the default texture coordinates for a simple box shaped IndexedFaceSet with an X dimension twice as large as the Z dimension and four times as large as the Y dimension. Figure 13.7 illustrates the original texture image used on the IndexedFaceSet used in Figure 13.6.

The current result is that both S and T coordinates are stretch to the [0, 1] range while T should be between 0 and some S ratio.

Wrong (current implementation):
![capture d ecran 2014-09-02 a 14 51 21](https://cloud.githubusercontent.com/assets/866788/4118368/3f405c3e-32a0-11e4-86a3-7b094ad71aad.png)
Correct:
![textured_shapes](https://cloud.githubusercontent.com/assets/866788/4118367/3f3bd312-32a0-11e4-9ff7-9898ab37d0f2.png)

PR #749, opened 2017-07-05
Fix default UV mapping

fabienrohrer:
Fix https://github.com/x3dom/x3dom/issues/411

## Summary

The X3Dom default UV mapping is not respecting well the X3D reference:

http://www.web3d.org/documents/specifications/19775-1/V3.3/Part01/components/geometry3D.html#f-IndexedFaceSettextureDefaultMapping

The issue is that the second UV coordinate (aka `T`) should be truncated by the first coordinate (aka `S`).

## Test X3D file

[default-mapping.zip](https://github.com/x3dom/x3dom/files/1124198/default-mapping.zip)

## Results

Note: I tested on my implementation and on Blender giving the same results.

| Original | Patched | Reference (my implementation) |
| --- | --- | --- |
| capture d ecran 2017-07-05 a 11 40 48 | capture d ecran 2017-07-05 a 11 39 26 | capture d ecran 2017-07-05 a 11 39 04 |

### DIFF
--- a/src/Mesh.js
+++ b/src/Mesh.js
@@ -409,8 +409,7 @@ x3dom.Mesh.prototype.calcTexCoords = function(mode)
         for (var k=0, l=0, m=this._positions[0].length; k<m; k+=3)
         {
             this._texCoords[0][l++] = (this._positions[0][k+S] - sMin) / sDenom;
-            this._texCoords[0][l++] = (this._positions[0][k+T] - tMin) / tDenom;
+            this._texCoords[0][l++] = (this._positions[0][k+T] - tMin) / sDenom;
         }
     }
 };
-


andreasplesch:
It would be great if you could test against (some of) these conformance examples:

http://www.web3d.org/x3d/content/examples/ConformanceNist/Appearance/ImageTexture/index.html
and perhaps also these:
http://www.web3d.org/x3d/content/examples/ConformanceNist/Appearance/TextureTransform/index.html

Not sure if these examples actually cover the truncation testing but they should keep working in any case.

[With User scripts via tampermonkey or greasemonkey, it is possible to replace the x3dom.js on the page with a fixed one; it may suffice to just make the fixed Mesh.js a User script].

fabienrohrer:
Thank you, I didn't know these tests. However I used my own tests instead of the monkey scripts.

I observed the conformance examples, and the best one showing the issue is `indexedfaceset`:

http://www.web3d.org/x3d/content/examples/ConformanceNist/Appearance/ImageTexture/IndexedFaceSetIndex.html

In this case, the IndexedFaceSet is defined without any texture coordinate. Here are the results:

conformance

Without patch, the texture is wrongly stretch on the entire surface of the parallelepiped, while with the patch, as in the conformance test, the texture height is correctly cut on this face.
=> I added white arrows to indicate you where to read this value.

fabienrohrer:
Do you want more tests? Honestly, I'm sure my patch is correct :-)

andreasplesch:
LGTM, also see as reference:
http://www.web3d.org/x3d/content/examples/ConformanceNist/Appearance/ImageTexture/_viewpoints/IndexedFaceSet.x3d._VP_Default_viewpoint.png

 You will need to ping @tsturm or @mlimper to accept the PR.

fabienrohrer:
> Do the other examples also look ok ?

I didn't check much of the other examples, I can try to use the monkey scripts as you described earlier, but I fear to spend a lot of time for this.
Would it be possible to delegate the tests to someone having more knowledge than me in the X3Dom workflow? If not, I can give a try.


> You will need to ping @tsturm or @mlimper to accept the PR.

How can I do this? If they are mentioned in this PR, they should be noticed, right?

andreasplesch:
A quick way is to make a gist:
https://gist.github.com/andreasplesch/19ead05b94c8a3482a086828d20b469e
The filename has to end with .user.js
and there needs to be the User script meta data block as comments at the start.
The just clicking on raw will trigger installation of the User script.
I went through some of the conformance example and there was not anything obviously wrong.

Just including @tsturm and @mlimper should trigger a notification, yes.

fabienrohrer:
Thank you a lot, I never used gist before, and this is clear enough now. However, I'm still lost about the user script. Is there some documentation somewhere on what it is and how to use it? 

https://github.com/x3dom/x3dom/wiki/Testing-your-changes is empty 😭

fabienrohrer:
If I understand well this, I have other patches to submit ;-)

andreasplesch:
https://addons.mozilla.org/en-US/firefox/addon/greasemonkey/

Or tampermonkey for chrome

https://github.com/OpenUserJs/OpenUserJS.org/wiki/Userscript-beginners-HOWTO

fabienrohrer:
@andreasplesch Thank you a lot, this is extremely useful and interesting.

So..
I used your gist for the following tests. Here is a serie of screenshots showing the results:
- On top: the Cobweb reference
- On bottom-left: the current X3Dom version
- On bottom-right: the patched X3Dom version

1. http://www.web3d.org/x3d/content/examples/ConformanceNist/Appearance/ImageTexture/IndexedFaceSetIndex.html

(The patch is fixing the default UV-mapping of the IndexedFaceSet)

capture d ecran 2017-07-10 a 16 04 16


2. http://www.web3d.org/x3d/content/examples/ConformanceNist/Geometry/Box/textureX3dom.xhtml

(The patch is not causing side effect on the other default UV-mapping, here the box.)

capture d ecran 2017-07-10 a 16 01 58

fabienrohrer:
Thanks a lot :-)

Issue #761, opened 2017-08-05
CesiumMilkTruck wheels redux for MC


andreasplesch:
https://rawgit.com/andreasplesch/gltf-test/addLight/examples/x3dom/index.html?category=sampleModels&model=CesiumMilkTruck&scale=0.5&type=glTF-Binary-MaterialsCommon

see #709
 
the fix is the same but needs to be hooked into KHRMaterialCommons.bind() in addition to glTFMaterial.bind()

andreasplesch:
fixed by #762

PR #759, opened 2017-08-05
gfTF khr material commons transparency/specular fix

andreasplesch:
https://github.com/KhronosGroup/glTF/tree/master/extensions/Khronos/KHR_materials_common defines the transparency property as providing opacity values, with a default of 1.0 for no transparency. All official glTF examples use the default.
https://rawgit.com/andreasplesch/gltf-test/Binary-MC/index.html provides the models. Without the transparency fix all models are displayed as completely transparent.
https://rawgit.com/andreasplesch/gltf-test/addLight/index.html applies the fixed shader from this PR to all example models. With it the models are displayed.
There are other issues with specular colors and node transforms which may be addressed separately.
This PR addresses #758.

PR #762, opened 2017-08-05
glTF kmc trafos fix

andreasplesch:
glTTF nodes using a KHR Material Commons material also needed to have their world transform applied, in the same ways as #723 does for generic glTF material nodes.
This addresses #761 .
https://rawgit.com/andreasplesch/gltf-test/kmcTransforms/examples/x3dom/index.html?category=sampleModels&model=CesiumMilkTruck&scale=0.5&type=glTF-Binary-MaterialsCommon
uses this fix (and #759)
https://rawgit.com/andreasplesch/gltf-test/kmcTransforms/index.html provides viewing of all other sample models with this patch (and #759).


andreasplesch:
PR build tested

PR #767, opened 2017-09-03
CoordinateDouble node implementation

andreasplesch:
This node may help with scene compatibility across x3d browser. Example use: https://jsfiddle.net/fzzgso28/1/
Also see http://web3d.org/pipermail/x3d-public_web3d.org/2017-September/007421.html and thread for some discussion.
The CoordinateDouble.js file was placed next to Coordinate.js file in the Rendering folder although it belongs with Nurbs component. Feel free to move to another location.


andreasplesch:
Since this is a new file, it needs to be added to tools/package.json for building.

PR #768, opened 2017-09-03
[Event utilities]Integer/BooleanSequencer

andreasplesch:
These are implementations of the abstract X3DSequencerNode, the IntegerSequencer, and BooleanSequencer implementations, which are part of the Event Utilities component.
The implementation is based on the X3DInterpolatorNode code.
https://x3dom-sequencers.glitch.me/ and
https://x3dom-sequencers-next.glitch.me/
exercise these nodes.


andreasplesch:
Since these are new files in a new component, they need to be added in a new EventUtilities component to tools/packages.json for building.

andreasplesch:
moved to #770

PR #775, opened 2017-10-05
Add property to NavigationInfo for natural Scrolling. Fixes #486

### LINKED ISSUE #486: Mouse wheel zoom behaviour is opposite to some other players
In "examine" mode, scrolling the mouse wheel up, zooms out. This is exactly opposite to the convention used by almost all other applications. Web browsers for instance, zoom in (enlarge) with <ctrl> wheel scroll up.

frattay:
This PR adds a property called `naturalScoll` to the NavigationInfo-Node, which inverts the scroll-events and thus fixes #486.

### DIFF
--- a/src/X3DCanvas.js
+++ b/src/X3DCanvas.js
@@ -293,8 +293,12 @@ x3dom.X3DCanvas.prototype.bindEventListeners = function() {
             this.focus();
 
             var originalY = this.parent.mousePosition(evt).y;
-
-            this.mouse_drag_y += 2 * evt.detail;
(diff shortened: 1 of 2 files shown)


andreasplesch:
Hopefully something like this can get merged.

I think this PR changes both zoom behaviours wheel and mouse dragging. From #486 it looked most would only like to change wheel behaviour ?
 
Perhaps make it a scalar factor: SFField wheelSpeed with a default of -2, to allow sensitivity adjustments ?

'natural' may be too opinionated but this is just my opinion.

andreasplesch:
tested, merging

PR #780, opened 2017-10-21
Let <code> elements wrap white space

paulmasson:
Earlier today I tried reading your documentation page for IndexedFaceSet on my iPad. The entire page was zoomed out to accommodate the &lt;code&gt; element near the top, making it impossible to read without zooming in. Even for browsers that do not zoom out for extremely long lines, most of the text will be outside of the visible window and not easily accessible.

This small change keeps the HTML encoding of the node and all of its defaults in a compact wrapped area where everything is easily seen, making the documentation more useful.

### DIFF
--- a/doc/x3doc/base/static/styles/x3dom-docs-new.css
+++ b/doc/x3doc/base/static/styles/x3dom-docs-new.css
@@ -697,7 +697,7 @@ code {
     font-size: 90%;
     color: #c7254e;
     background-color: #f9f2f4;
-    white-space: nowrap;
+    white-space: normal;
     border-radius: 0;
 }
 kbd {


andreasplesch:
a screenshot on desktop chrome with white space: normal for <code>:
image

andreasplesch:
and using the device preview developer tools option on chrome:
image

andreasplesch:
So the PR would look good to me, eg. an improvement. If there are other doc style improvements for mobile/tablets, please feel free to submit another PR.

andreasplesch:
Any objections ?

andreasplesch:
It looks like the documentation on x3dom.org still needs to be regenerated, eg, generation is not part of the nightly updates. Hopefully somebody with access can trigger this.

paulmasson:
@andreasplesch The documentation has still not been regenerated. Any idea who can do this? Thanks.

andreasplesch:
I noticed and was hoping that somebody at FHG would get around doing that ...

PR #782, opened 2017-10-25
fix boolean_changed in routes

andreasplesch:


andreasplesch:
merge and rebase worked fine

PR #801, opened 2017-12-12
Typos

paulmasson:


andreasplesch:
Thanks !

paulmasson:
@andreasplesch I wasn't sure if these would be considered typos or not, since they work either way with the current implementation. 

Is there an expectation for the future that node names will insist on a particular capitalization, like with SVG files, or is this expected to remain a matter of style?

andreasplesch:
In HTML documents, element names are not case sensitive in general. So both
the X3D official capitalized names and lowercase names work. In XHTML
documents element names are case sensitive and only the official names are
recognized.




On Dec 12, 2017 4:15 PM, "Paul Masson" <notifications@github.com> wrote:

@andreasplesch <https://github.com/andreasplesch> I wasn't sure if these
would be considered typos or not, since they work either way with the
current implementation.

Is there an expectation for the future that node names will insist on a
particular capitalization, like with SVG files, or is this expected to
remain a matter of style?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/x3dom/x3dom/pull/801#issuecomment-351196772>, or mute
the thread

PR #802, opened 2017-12-12
Update X3DVolumeDataNode.js

rbakker:
In the function texture3DFunctionShaderText, the lines dy1 = floor(s1/nX)/nY; and dy2 = floor(s2/nX)/nY; are wrong, one must divided by nX instead of nY to get the correct texture position. Apparently this has been tested only on square texture-grids where nX equals nY.
This bug affects most volume rendering nodes!

### DIFF
--- a/src/nodes/VolumeRendering/X3DVolumeDataNode.js
+++ b/src/nodes/VolumeRendering/X3DVolumeDataNode.js
@@ -166,9 +166,9 @@ x3dom.registerNodeType(
                 "  s1 = floor(volpos.z*nS);\n"+
                 "  s2 = s1+1.0;\n"+
                 "  dx1 = fract(s1/nX);\n"+
-                "  dy1 = floor(s1/nY)/nY;\n"+
+                "  dy1 = floor(s1/nX)/nY;\n"+
                 "  dx2 = fract(s2/nX);\n"+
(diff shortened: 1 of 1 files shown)


andreasplesch:
@aarbelaiz ? Do you have a test scene for which the fix is necessary ? We will also have to make sure that the example will continue to work.

aarbelaiz:
This is a good catch, the proposed fix is correct and should be merged. It will not break any current examples. 
I always compose texture atlases in square textures, @rbakker do you have a test scene? otherwise I would create one as soon as I can.

rbakker:
I created a test scene by taking the 10x10 aorta and changing it into 7x14.
The bug is displayed here:
https://scalablebrainatlas.incf.org/x3bug/aorta_c7r14.xhtml
And fixed here:
https://scalablebrainatlas.incf.org/x3bug/aorta_c7r14_fixed.xhtml

andreasplesch:
I (travis) build a package with the patch and put it here
http://x3dom-builds.surge.sh/latest/

The examples at
http://examples.x3dom.org/volren/volrenOpacityTestTF_aorta.xhtml
http://examples.x3dom.org/volren/volrenMPR_aorta.xhtml

work fine with it. (I use https://x3dom-test.glitch.me/ to test)

So this all looks good to me.

PR #803, opened 2017-12-12
fix parentRemoved function

rbakker:
When removing a VolumeData node from an X3D scene, the volume isn't removed and a console error is generated ':
Uncaught TypeError: Cannot set property 'texture' of undefined. See this example:
https://scalablebrainatlas.incf.org/x3bug/aorta_remove.xhtml
The added check in line 191 fixes the problem. I do not know what is really causing the error, whether more elegant solutions exist, and whether perhaps a similar fix is needed in line 175. The following page illustrates the desired behavior after the fix:
https://scalablebrainatlas.incf.org/x3bug/aorta_remove_fixed.xhtml

### DIFF
--- a/src/nodes/Texturing/X3DTextureNode.js
+++ b/src/nodes/Texturing/X3DTextureNode.js
@@ -188,7 +188,9 @@ x3dom.registerNodeType(
                     else {
                         // Texture maybe in MultiTexture or CommonSurfaceShader
                         Array.forEach(shape._parentNodes, function (realShape) {
-                            realShape._dirty.texture = true;
+                            if (x3dom.isa(realShape, x3dom.nodeTypes.Shape)) {
+                                realShape._dirty.texture=true;
+                            }
                         });
                     }
                 });
--- a/src/nodes/Texturing/X3DTextureNode.js
+++ b/src/nodes/Texturing/X3DTextureNode.js
@@ -181,16 +181,14 @@ x3dom.registerNodeType(
             parentRemoved: function(parent)
             {
                 Array.forEach(parent._parentNodes, function (shape) {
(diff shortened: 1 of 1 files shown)


andreasplesch:
While having this guard seems safe, it would be good to figure out what the underlying problem is.

If first the VolumeData element is removed, and then the transform there is no error.

I think the problem has do with how VolumeData is a X3DShapeNode with 'builtin' Appearance and Texture nodes. 

The error occurs when OpacityMapVolumeStyle is the parentRemoved of the texture. parent._parentNodes[0] here is the VolumeData node.

```js
parentRemoved: function(parent)
            {
                Array.forEach(parent._parentNodes, function (shape) {
                    // THINKABOUTME: this is a bit ugly, cleanup more generically
                    if (x3dom.isa(shape, x3dom.nodeTypes.X3DShapeNode)) { // allow VolumeData
                        shape._dirty.texture = true;
                    }
```

therefore would be more correct if it works. Could you please test this ? Thanks.
(Similar may apply throughout ..)

rbakker:
Hi Andreas, your proposed patch works for me. I have implemented it in x3dom.debug.js on line 47783 and updated the test page to use the patched script:
https://scalablebrainatlas.incf.org/x3bug/aorta_remove_fixed.xhtml

andreasplesch:
Let's test this a bit more but looks good to me.

andreasplesch:
http://x3dom-builds.surge.sh/build_6/x3dom-full.debug.js is now the build with this patch applied. Testing revealed an unrelated problem with removing a VolumeData element with RadarVolumeStyle (https://examples.x3dom.org/example/RadarVolumeStyle/) due to RenderedTexture.

PR #804, opened 2017-12-14
Avoid CORS error

paulmasson:
This image produces a CORS redirection error in Safari, but not in Chrome or Firefox. This simple edit solves that particular problem.


andreasplesch:
Thanks !

PR #805, opened 2017-12-15
Correct default orientation

paulmasson:
I noticed a discrepancy in the documentation for this default value:

http://www.web3d.org/documents/specifications/19775-1/V3.3/Part01/components/navigation.html#Viewpoint

This was presumably a simple transposition typo, since a rotation about a zero axis makes no sense.

### DIFF
--- a/src/nodes/Navigation/Viewpoint.js
+++ b/src/nodes/Navigation/Viewpoint.js
@@ -57,11 +57,11 @@ x3dom.registerNodeType(
              * The orientation fields of the Viewpoint node specifies relative orientation to the default orientation.
              * @var {x3dom.fields.SFRotation} orientation
              * @memberof x3dom.nodeTypes.Viewpoint
-             * @initvalue 0,0,0,1
+             * @initvalue 0,0,1,0
              * @field x3d
              * @instance
              */
-            this.addField_SFRotation(ctx, 'orientation', 0, 0, 0, 1);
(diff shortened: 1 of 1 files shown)


andreasplesch:
Yes, 0 0 1 0 would be the correct default. For the nodes the documentation should be generated from the code but I do not know if this is still happening. So it would be best to correct Viewpoint.js as well which should not lead to complications.

paulmasson:
The changes were made in Viewpoint.js originally.

andreasplesch:
Yes, of course.

paulmasson:
Since this is a straightforward correction of an obvious error in the code, what is the protocol for getting this pull request merged?

Also, the certificate on the x3dom.org website has expired and is producing security errors. What is the best way to contact the person responsible?

andreasplesch:
The error is benign since the weird orientation gets immediately converted during parsing to a quaternion with the correct value. But this is still a bug, We will just need to actually make sure with quite a few test scenes that the default view continues to work 100% since it is wide use. It is a bit of a sensitive change requiring some extra care although I do not anticipate problems.

andreasplesch:
http://x3dom-builds.surge.sh/build_7/ has builds with the pr patch applied for testing. I could not find any problems but lets give it a bit more time.

andreasplesch:
Well, I tested quite a few scenes, without problems. I still feel there may have been some motivation originally but we will see.

PR #828, opened 2018-03-13
Remove a couple hard-coded dates in documentation

paulmasson:
The code for this change was copied from the template doc/x3doc/tmpl/layout.tmpl


paulmasson:
This has no effect. Please ignore.

PR #829, opened 2018-03-17
Remove reference to generation

paulmasson:
These two documentation files are both static, so the reference to generation is misleading. In the event that changes in the future, I just commented out the two references.


andreasplesch:
Thanks !

PR #831, opened 2018-03-19
JSON Prototypeexpander...for expanding prototypes into non-prototypes.   Does not work with scripts.  Duplicates nodes.  Alternative pull request than #830

coderextreme:
Added to JSON Parser pull request.  Apply that JSON Parser pull request first, otherwise, this will parse JSON files in Inline.js

Issue #688, opened 2016-11-30
Track Downloads per Multipart Node


mlimper:

bogdanelloo:
I am already using an slightly extended version of DownloadManager and a x3dom implementation with respect to the previously mentioned DM in order to have full control on all XHR calls in the executionqueue such that I can cancel them on geom/mesh change event => increased speed, lower bandwidth, better UX and it is working as a charm!

mlimper:
That sounds great - looking forward to your PR!

david-zacharias:
@bogdanelloo Please share more information about your extended version.

bogdanelloo:
Hi,

I have checked with our legal dept. and it seems that we need some approval in order for me to be able to post the sources here.

While waiting for the approval:
@mlimper - I can send it to you over e-mail.
@david-zacharias - I cannot share the full source at this moment, but I would be happy to answer any specific questions that you have.

mlimper:
> @mlimper - I can send it to you over e-mail.

You could, but then I am not sure how we would proceed - our goals would anyway be to have this functionality in the main X3DOM repository. Another way would be to set up NDAs to share this feature "offline" with selected partners - but this is against the spirit of X3DOM, and also makes things pretty complicated.

Maybe we can discuss at a higher level, without source code, while waiting for approval.
So, here's a more specific question, related to this ticket:

Is your current implementation able to tell how many downloads are currently left for a specific MultiPart? Imagine, for example, your scene would consist of three major components, each in one MultiPart, and you would like to see, for each MultiPart, when its downloads have been finished. This is the essential feature we are looking for right now.

bogdanelloo:
@mlimper 

> Is your current implementation able to tell how many downloads are currently left for a specific MultiPart? 
No, it is not!
I am currently tracking all async XHR calls in the execution queue without diff. with respect to a specific multi-part.

tsturm:
I inspect the code and the DownloadManager was only used for the PopGeometry, not for BinaryGeometry and not for anything else.

To not break the PopGeometry I added the new RequestManager, which now handles all x3dom downloads with the ecxeption of the PopGeometry.

With the new RequestManager you are able to:

* Set the maximum number of parallel requests:

    ```x3dom.RequestManager.maxParallelRequests```

* Set the withCredentials flag for all requests:

    ```x3dom.RequestManager.withCredentials```

* Add request headers for all requests:

   ```x3dom.RequestManager.addRequestHeader( header, value );```

[...]

PR #837, opened 2018-04-08
Fixed per comments on pull request #831. Replacement for pull request #831.  Now includes ExternProto expander

coderextreme:


coderextreme:
It’s in src/util/json/PrototypeExpander.js (grab most recent).  I test for $ before calling $.ajaxSetup or $.get.  I have a need for the ExternPrototypeExpander to work on clients in X3DOM, but on servers in X3DJSONLD (for testing and debugging purposes), so we can get rid of the Extern code, but that means X3DOM is less functional without a server, and I have to maintain 2 code bases (which I do anyway—3 for JSONParser—trying to make things simpler).  I *just* integrated the Extern expander, and would like to keep it in my code.   Thus I need $.get or something.

I’ve got a pure get somewhere that uses promises, but I still need to use $.ajaxSetup to set async false.  Got a fix for that?

So a .get that works on both client and server, that works with relative and absolute URLs would be most desirable.

As far as I can see x3dom and JQuery appear to be somewhat compatible, since I use them together a lot. Why do you see it as a blocker?  I agree that we should not depend on JQuery and I will look for another answer.  I have a get in my HTML code in X3DJSONLD.   I also use $ in X3DJSONLD.js when I shouldn’t (it’s not designed to be dependent on JQuery), I just would like something open source and easy in my base libraries for X3DJSONLD that I don’t have to maintain or really think about.  However, I may want to start using promises, and I don’t think $.get works too well with promises.  Basically, I want something that will load a file from the file system or the web server without much fuss.  If there was something to replace loadURLs and processURLs I would use it if I knew what it was.

Is there a good example of using x3dom requestmanager?  Perhaps I can start using X3DOM more integrated into X3DJSONLD, instead of as an external dependency (with my own build).

I probably also need to integrate the DOM2JSONSerializer for XML extern protos.   More work.

John

Sent from Mail for Windows 10

andreasplesch:
#688 introduces RequestManager. It is used for Inline, and applications.

You can set async for XHR:

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests

Could you run a (express or python) server on the server to serve files to local x3dom on the server ?

x3dom is already too heavy and should not require more dependencies.

coderextreme:
Yes, but that doesn’t work on my webhost which only supports PHP for the most part. Will have to write something in PHP eventually.  I like client side and I got it to work with sync XHR.

John

Sent from Mail for Windows 10

coderextreme:
I have removed the dependency on jquery, even though I never planned to require that all X3DOM users use it.  The code is much better, and it’s pretty easy to do synchronous xhr.

John

Sent from Mail for Windows 10

src/nodes/Core/X3DNode.js
```diff
@@ -293,6 +293,9 @@ x3dom.registerNodeType(
                                 case "string":
                                     this._vf[field] = msg;
                                     break;
+				default:
```
andreasplesch: not sure about this default. How is it necessary ?

src/util/json/PrototypeExpander.js
```diff
@@ -0,0 +1,1229 @@
...
+				
+		}
+		// console.error("Processed URLs", localArray.join(" "));
+		return localArray;
+	},
+
+	/**
+	 * Use almost any method possible to load a set of URLs.  The loadpath is the
+	 * original URL the main JSON got laoded from, Urls is the se of urls, and
+	 * the loadedCallback returns the data and the URL it was loaded from.
```
andreasplesch: Perhaps add that the http(s) and fs modules are required on server with node.js ?

andreasplesch:
Here are test builds with the latest commits merged:

http://x3dom-builds.surge.sh/build_5/
http://x3dom-builds.surge.sh/build_5/x3dom-full.debug.js

In fact, each time commits are added to this PR, a new build is triggered automatically and pushed to http://x3dom-builds.surge.sh

coderextreme:
I removed the default case and added comments for loadURLs on the server

andreasplesch:
Here is a test page to load json inlines:
https://x3dom-json-inline-test.glitch.me/index.xhtml

andreasplesch:
I had to revert since DataUrls are not correctly recognized, eg. will be assumed .json. They could be either .x3d or .json, once decoded . The mime type for both would be text/plain . Not sure. May be check how x_ite distinguishes between .x3d and .wrl for DataUrls.

coderextreme:
I believe x_ite distinguishes by trying to parse and failing.

John.

src/nodes/Networking/Inline.js
```diff
@@ -159,9 +159,23 @@ x3dom.registerNodeType(
...
-                    xhr.overrideMimeType('text/xml');   //application/xhtml+xml
+                if (this._vf.url.length && this._vf.url[0].length) {
+			if (this._vf.url[0].toLowerCase().endsWith(".x3d")) {
+				isJSON = false;
+				if (xhr.overrideMimeType)
+				    xhr.overrideMimeType('text/xml');   //application/xhtml+xml
+			} else if (this._vf.url[0].toLowerCase().endsWith(".json")) {
+				if (xhr.overrideMimeType)
+				    xhr.overrideMimeType('application/json');
+			}
```
andreasplesch: handle other cases like DataURL

src/nodes/Networking/Inline.js
```diff
@@ -159,9 +159,23 @@ x3dom.registerNodeType(
               
                 var that = this;
 
+		var isJSON = true;
```
andreasplesch: make default isJSON = false for backwards compatibililty

src/nodes/Networking/Inline.js
```diff
@@ -159,9 +159,23 @@ x3dom.registerNodeType(
...
                 var xhr = new window.XMLHttpRequest();
-                if (xhr.overrideMimeType)
-                    xhr.overrideMimeType('text/xml');   //application/xhtml+xml
+                if (this._vf.url.length && this._vf.url[0].length) {
+			if (this._vf.url[0].toLowerCase().endsWith(".x3d")) {
+				isJSON = false;
+				if (xhr.overrideMimeType)
+				    xhr.overrideMimeType('text/xml');   //application/xhtml+xml
+			} else if (this._vf.url[0].toLowerCase().endsWith(".json")) {
+				if (xhr.overrideMimeType)
```
andreasplesch: isJSON -= true;

src/nodes/Networking/Inline.js
andreasplesch: for now, could just assume x3d as it was the existing behaviour

coderextreme:
I issued another pull request patching the isJSON, and hopefully restoring data urls,  please test with inline data urls and x3d.  Thanks!

John

Sent from Mail for Windows 10

PR #860, opened 2018-05-28
Browser sai

coderextreme:


coderextreme:
Just fixed comments.  Haven't verified previous merge from core repo works (it doesn't, looking into it).

coderextreme:
Added loadJS() and my stuff started working again after mods.  I am not quite sure why, but I was calling different stuff on my side (working with DOM instead of JS objects), so that might have been it.  Perhaps something was cached?

PR #863, opened 2018-06-04
Browser json sai

andreasplesch:


andreasplesch:
@coderextreme : I want to wrap this. There is issue I discovered. Since replaceWorld() is now replacing not just the scene but the complete x3d element (but preserving its attributes, eg. size and position) - in order to use .reload() which is cleaner -, a new .runtime object is also created. This means it is necessary to retrieve this new runtime each time after a .replaceScene (and all other methods using it). This is easy to do by just reassigning (`runtime = ele.runtime`) but perhaps not expected since the spec. implies that you could reuse the same Browser instance. I think it would be still ok and just needs to be documented somewhere, probably in a comment at the code for now. What do you think ?

coderextreme:
Did you want me to test something?   Or is there extra code to add outside
the added calls to x3dom?

coderextreme:
It seems like if you are replacing the runtime, I should not keep a
reference to it.   Which I don't.   I believe I use doc.runtime between
calls.

If you want to add comments it should be ok.   Basically it sounds like
people should retrieve the X3D element each time.
John

andreasplesch:
Yes, please test .replaceWorld() with using element.runtime.showAll('posX') after replacing. You can use
http://x3dom-builds.surge.sh/latest/
To make this work you will need to dynamically refind the x3d element each time after replace, for example by document.querySelector('x3d').runtime . Reusing an initially assigned x3delement should not work.

andreasplesch:
oh, I did not see the last meesage. So, yes, exactly. I will add a comment.

PR #867, opened 2018-07-19
HAnim cpu skinning

andreasplesch:
This PR adds basic skinning support for HAnim, along some other small fixes. The skin deformation is done during traversing in CollectDrawables in Joint, with js on the cpu. It works reasonably well for simple geometries:
https://rawgit.com/andreasplesch/x3dom/HAnim_work/test/functional/HAnim/inlineloader.html
JoeKick, Lucy and BoxMan use animated skinning.
The vertex deformation will need to move to the shader, perhaps using this as a starting point.


andreasplesch:
Anybody can of course give feedback, review. I will let this PR sit for a while. It should not have any impact outside of HAnim.

PR #877, opened 2018-08-05
Lint & formatting of Bindable.js -> Internals.js

jamesleesaunders:
Standardise formatting, parenthesis, comments and JSDoc across files Bindable.js to Internals.js

These files had a right mixed bag of formatting and indenting!
* Added JSDoc headers to all functions.
* Indenting 4 spaces, no tabs.
* Collapsed if / else statements.

I don't know if this is a bit too big a change as this diff is quite large. 
Do let me know if you would prefer I went a little slower or broke this down into smaller PR's.

### DIFF
--- a/src/Bindable.js
+++ b/src/Bindable.js
@@ -1,4 +1,4 @@
-/*
+/**
  * X3DOM JavaScript Library
  * http://www.x3dom.org
  *
@@ -9,11 +9,17 @@
  * Philip Taylor: http://philip.html5.org
  */
 
-
-///////////////////////////////////////////////////////////////////////////////
-// BindableStack constructor
-///////////////////////////////////////////////////////////////////////////////
-x3dom.BindableStack = function (doc, type, defaultType, getter) {
+/**
+ * BindableStack constructor
+ *
+ * @param doc
+ * @param type
+ * @param defaultType
+ * @param getter
+ *
+ * @constructor
+ */
+x3dom.BindableStack = function(doc, type, defaultType, getter) {
     this._doc = doc;
     this._type = type;
     this._defaultType = defaultType;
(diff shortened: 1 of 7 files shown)


andreasplesch:
Did you have a chance to go through tests ? I know this is just formatting and commenting but it still touches all sensitive core pieces. So it will be necessary to check as many (easy to run) tests as possible.

Smaller PR will be easier to quickly review. A PR per file is what I would recommend for this.

When you are done, you may be the only person ever who looked at each line of code, however briefly ;)

jamesleesaunders:
Sure - I will try to keep the PR's down to one file (this was my intention but I got carried away with this one!). 

Tests... I have run though all the html tests in the /tests/functional/ folder and most appear to work (a few of them spit console errors in Chrome (but this is also true for the master branch so no different with my changes). 
Apart from manually running down all the html files and running all benchmark tests is there a more automated test suite I should also be running?
I have also tested build and this works.

Also while running benchmark/field_test.html I get console error:
`x3dom.SFVec3f is not a constructor`
This is still true of master.  Believe this should be `x3dom.fields.SFVec3f` ? I can raise a separate PR/issue (if you confirm).

andreasplesch:
https://github.com/x3dom/x3dom/tree/master/test/regression-suite/test/cases

has another set of tests. There used be some automation for those but I believe it is not functional any more. Just running them though manually is where it is. With rawgit.com it is relatively quick to do that. Most of the tests with textures will need a crossOrigin='anonymous' attribute for the texture nodes to work since they are from a time when cors was not enforced by browsers.

The  benchmark/field_test.html could be updated but I think it has not been used in a long time.

jamesleesaunders:
Thanks. I have run down a good number of the tests in regression-suite folder and all look fine.
I have not manually run every one of them (there are a lot of them!).

PR #900, opened 2018-09-01
BufferGeometry Renaming

### LINKED ISSUE #898: glTF: rename buffergeometryaccessor  to bufferaccessor
Since buffers are used not just for geometry but also other data such as animation data, renaming the buffergeometryaccessor/view nodes to buffer accessor/view would allow use of those more generally.

I can work on a PR if desired.

jamesleesaunders:
Closes #898. 

> Since buffers are used not just for geometry but also other data such as animation data, renaming the BufferGeometryAccessor/View nodes to bufferAccessor/View would allow use of those more generally.

This is just a find and replace renaming of occurrences of
* `BufferGeometryView` to `BufferView`
* `buffergeometryview` to `bufferview`
* `BufferGeometryAccessor` to `BufferAccessor`
* `buffergeometryaccessor` to `bufferaccessor`

Tested various regression-suite tests and the run fine.

As the intention is to re-use the buffer for more than just geometry, I was wondering if you also wanted these files moving out of the `nodes/Geometry3D/` folder? Possibly into a new `nodes/Buffer/` folder? Or into `nodes/Core/`?

### DIFF
--- a/src/gfx_webgl.js
+++ b/src/gfx_webgl.js
@@ -313,7 +313,8 @@ x3dom.gfx_webgl = (function() {
             // Check if we need a new shader
             shape._webgl.shader = this.cache.getShaderByProperties(gl, shape, shape.getShaderProperties(viewarea));
 
-            if (!needFullReInit && shape._webgl.binaryGeometry == 0 && shape._webgl.bufferGeometry == 0)    // THINKABOUTME: What about PopGeo & Co.? {
+            if (!needFullReInit && shape._webgl.binaryGeometry == 0 && shape._webgl.bufferGeometry == 0) {
+                // THINKABOUTME: What about PopGeo & Co.?
                 for (q = 0; q < shape._webgl.positions.length; q++) {
                     q6 = 6 * q;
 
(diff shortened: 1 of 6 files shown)


andreasplesch:
I think BufferGeometry should not be renamed to Buffer. It is actually a Geometry3D. Buffer is a more general glTF term for gl data storage.

The folders correspond to the X3D components:
https://www.web3d.org/files/specifications/19775-1/V3.3/Part01/Architecture.html

All the binary, external geometry nodes should actually be in Geometry3DExt until they are adopted into the standard but traditionally they were put into the Geometry3D. I think the reason may have been to bundle them with x3dom.js, and not just x3dom-full.js.

I think Networking may be the closest fit for now for the BufferView/Accessor nodes but not ideal either. Let's just keep them with BufferGeometry.js, in Geometry3D. @tsturm ?

jamesleesaunders:
No worries - all done. Reverted BufferGeometry -> Buffer renaming. Only renamed BufferGeometryAccessor and BufferGeometryView. Leaving files in Geometry3D directory.

andreasplesch:
I tested inline-gltf.html with the Khronos repo:
```js
            var repo = "https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/";
```
and it works fine. LGTM.

PR #906, opened 2018-10-03
Take pixel ratio into account when current dimension is set

ksokolowski7:
Device pixel ratio was ignored when current dimension was set. It caused performance issue on point view for mobile devices and crashes on Safari.

### DIFF
--- a/src/X3DCanvas.js
+++ b/src/X3DCanvas.js
@@ -73,6 +73,7 @@ x3dom.X3DCanvas = function(x3dElem, canvasIdx)
                 if (that.doc && that.doc._viewarea) {
                     that.doc._viewarea._width = parseInt(that.canvas.getAttribute("width"), 0);
                     that.doc.needRender = true;
+                    that._current_dim[0] = that.doc._viewarea._width;
                 }
                 break;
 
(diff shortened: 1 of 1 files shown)


andreasplesch:
Is there a way to reproduce the issue on Chrome, FIrefox on desktop ?
Or is it restricted to safari on iphone ? Is there an example scene ? Is full screen styling affected ?
In addition to setting current dim in the setAttribute handler, canvas.style.width/height should also be updated. Not sure if there is a reason why it is not, currently. Perhaps setting canvas attributes used to suffice to modify the actual dimensions.

ksokolowski7:
Sorry, I do not have example scene. I can reproduce it on Chrome in mobile view. There are 2 errors:

image

I have option _backend_ set on "desktop" - not sure if its important information. 

It works on Chrome but there is worse performance. Safari (mobile) crashes in that time. 

In my case I open scene in dialog box in full screen, it can be reason why width/height is updated.
I was trying to move my scene to simple view but looks ok then. 

Looks that canvas.style.width/height is updated in setAttribute method by this.__setAttribute(attrName, newVal).

andreasplesch:
So from this tracing it looks like the best way to resize after initial parsing is to use the style of the attached canvas as in:

```
x3delement.runtime.getCanvas().style.width='100px';
```

Can you try this in Safari and see if that also crashes ?

andreasplesch:
As the PR fixes a crash (on Safari), it should be applied. But somebody will need to reproduce with a known example, and test a bit with other browser. Please feel free to reopen as needed.

PR #908, opened 2018-10-16
Texture does not load when authorization required for port different then 80 (Safari)

ksokolowski7:
CrossOrigin was set to "anonymous" when port was different then 80 on domain. It caused that cookie was not send with request about texture.

It did not work only for Safari.

### DIFF
--- a/src/util/Utils.js
+++ b/src/util/Utils.js
@@ -1293,7 +1293,7 @@ x3dom.Utils.forbiddenBySOP = function (uri_string) {
         Port = Host_Port[1];
     } // else will return false for an invalid URL or URL without authority
 
-    Port = Port || "80";
+    Port = Port || originPort;
     Host = Host || document.location.host;
     Scheme = Scheme || document.location.protocol;
     return !(Port === originPort && Host === document.location.host && Scheme === document.location.protocol);


ksokolowski7:
The whole page is served on port 8080. Object x3d is loaded from the same domain and the same port. The url to texture is inside x3d object and looks like "texture.png". So texture is under http://mydomain:8080/texture.png

My originPort for this case is 8080 as well. Then is compared with hardcoded port 80. 

It is different and result is that texture is loaded with crossOrigin anonymous (look on the line 422):

```
            default: {
                if(x3dom.Utils.forbiddenBySOP(src[i])) {
                    image.crossOrigin = 'anonymous';
                }
            }
```
I require additional authorization for each content and cookie is needed for texture as well. 

Why Safari only?
Looks that Chrome and FF send cookie with this example when domain and port is the same:
```
var image = new Image();
image.crossOrigin = "anonymous";
image.src = "test.png";
```

Safari does not send cookie when crossOrigin is set to anonymous.

andreasplesch:
Ok. "texture.png" is a relative URL and should therefore inherit the base document port:
http://www.ietf.org/rfc/rfc1808.txt?number=1808
Could you try the absolute URL for url field value as well ? url='http://mydomain:8080/texture.png'
It should work with and without the patch.

PR #926, opened 2019-01-29
dirty lighting by full array

andreasplesch:
rather than count, to allow for simultaneous adding and removing.
live at https://raw.githack.com/andreasplesch/x3dom/a2e5639963c50ef1218ba2174d0774aca37635fe/test/functional/inline-gltf.html
This change required the least modifications but there may be better ways.

### DIFF
--- a/src/util/Utils.js
+++ b/src/util/Utils.js
@@ -720,7 +720,7 @@ x3dom.Utils.isUnsignedType = function (str)
 *****************************************************************************/
 x3dom.Utils.checkDirtyLighting = function(viewarea)
 {
-	return (viewarea.getLights().length + viewarea._scene.getNavigationInfo()._vf.headlight);
+	return (viewarea.getLights().concat(viewarea._scene.getNavigationInfo()._vf.headlight));
 };
 
 /*****************************************************************************
--- a/src/util/Utils.js
+++ b/src/util/Utils.js
@@ -720,7 +720,15 @@ x3dom.Utils.isUnsignedType = function (str)
 *****************************************************************************/
 x3dom.Utils.checkDirtyLighting = function(viewarea)
 {
-	return (viewarea.getLights().concat(viewarea._scene.getNavigationInfo()._vf.headlight));
+	return (viewarea.getLights().length + viewarea._scene.getNavigationInfo()._vf.headlight);
+};
+
+/*****************************************************************************
+ * Checks for PhysicalEnvironmentLight change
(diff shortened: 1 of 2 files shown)


src/util/Utils.js
```diff
@@ -720,7 +720,7 @@ x3dom.Utils.isUnsignedType = function (str)
 *****************************************************************************/
 x3dom.Utils.checkDirtyLighting = function(viewarea)
 {
-	return (viewarea.getLights().length + viewarea._scene.getNavigationInfo()._vf.headlight);
+	return (viewarea.getLights().concat(viewarea._scene.getNavigationInfo()._vf.headlight));
```
tsturm: With this check all the shaders are regenerated every frame because it will compare two pointers to the same array.

Simultaneous adding and removing needs no shader regeneration.
The PhysicalEnvironmentLight is handled diffrent and needs an additional check.

andreasplesch: hm, right. There is the case that the env. light is removed and the headlight is added. This will result in no change in the count, and no regeneration although it is then needed, no ?

tsturm: I would add following function at line 736 in Utils.js

```
/*****************************************************************************
 * Checks for PhysicalEnvironmentLight change
 *****************************************************************************/
x3dom.Utils.checkDirtyPhysicalEnvironmentLight = function(viewarea, shaderProperties)
{
    return (!!shaderProperties.PHYSICALENVLIGHT == viewarea.hasPhysicalEnvironmentLight());
};
```

and change the getShaderProperties function in X3DShapeNode.js to:

```
getShaderProperties: function(viewarea)
{
    if (this._shaderProperties == null ||
        this._dirty.shader == true ||
        x3dom.Utils.checkDirtyEnvironment(viewarea, this._shaderProperties) ||
        x3dom.Utils.checkDirtyPhysicalEnvironmentLight (viewarea, this._shaderProperties) ||
        (this._webgl !== undefined &&
         this._webgl.dirtyLighting != x3dom.Utils.checkDirtyLighting(viewarea) ))
    {
        this._shaderProperties = x3dom.Utils.generateProperties(viewarea, this);
        this._dirty.shader = false;
        if (this._webgl !== undefined)
        {
             this._webgl.dirtyLighting = x3dom.Utils.checkDirtyLighting(viewarea);
        }
    }
    return this._shaderProperties;
},
```

andreasplesch: agreed and tested

PR #948, opened 2019-04-01
Added VR Controller orientation-based movement control

Joo918:
Hello.
My change checks the orientation of the left/right controllers the user is holding and if the user presses on button1, the viewarea moves in that direction.

### DIFF
--- a/src/util/VRControllerManager.js
+++ b/src/util/VRControllerManager.js
@@ -184,7 +184,14 @@ x3dom.VRControllerManager.prototype.onUpdate = function( viewarea, vrDisplay, co
 
         if(controllers.left.buttons[1].pressed)
         {
-            this.fit(viewarea);
+            //this.fit(viewarea);
+			var pose = controllers.left.pose;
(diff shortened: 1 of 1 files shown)


andreasplesch:
Thanks. VR navigation is still very experimental. Since the view all function ('a' key) is very useful in non immersive mono, we may want to preserve something similar for VR ?
There will be a need for compatible navigation with zero, one and two controllers, while also respecting Viewpoints. Hand gesture based navigation will also need to be considered.
So there will need to be a lot of experimentation at least as long as controller hardware is in flux.
It seems to me that user overrides for runtime hooks may be a way to allow for easy experimentation ?
AR may not need any navigation, just anchors with the real world.

Joo918:
I agree about the VR version of 'view-all' .
As for the compatibility of the number of controllers, my code runs only when button 1 of a VR controller is pressed (If that controller exists and is connected) and uses the orientation of that controller, so I believe there won't be a problem.
A support for navigation with no controllers should be considered, though.
I saw that VRControllerManager.js have support for axes-based navigation which could be used for controllers without orientation information such as a gamepad?

PR #959, opened 2019-05-25
fix layerXY for high DPR (mobile)

andreasplesch:
as discussed on the mailing list: https://sourceforge.net/p/x3dom/mailman/message/36676349/

### DIFF
--- a/src/Viewarea.js
+++ b/src/Viewarea.js
@@ -161,7 +161,7 @@ x3dom.Viewarea = function(document, scene) {
 
     this.vrControllerManager = new x3dom.VRControllerManager();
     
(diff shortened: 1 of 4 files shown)


src/Viewarea.js
```diff
@@ -160,6 +160,8 @@ x3dom.Viewarea = function(document, scene) {
     this.vrRightProjMatrix = new x3dom.fields.SFMatrix4f();
 
     this.vrControllerManager = new x3dom.VRControllerManager();
+    
+    this._inverseDevicePixelRatioInverse = 1.0 / window.devicePixelRatio;
```
tsturm: Is the double "inverse" in the variable name intended?

andreasplesch: thanks for the catch.

PR #965, opened 2019-06-04
Textrafo gltf mode

### LINKED ISSUE #963: TextureTransform center field regression
The NIST conformance scene for TextureTransform center field at

http://www.web3d.org/x3d/content/examples/ConformanceNist/Appearance/TextureTransform/center_primitivesX3dom.xhtml

does not look correct. It should look like this:

image

http://www.web3d.org/x3d/content/examples/ConformanceNist/Appearance/TextureTransform/_viewpoints/center_primitives.x3d._VP_Front_View.png

I am pretty sure that TextureTransform used to work correctly. yeah, 1.7.2 still works.

The NIST conformance scene for TextureTransform rotation still works fine:

http://www.web3d.org/x3d/content/examples/ConformanceNist/Appearance/TextureTransform/rotation_primitivesX3dom.xhtml

andreasplesch:
fixes #963

### DIFF
--- a/src/nodes/Texturing/MatrixTextureTransform.js
+++ b/src/nodes/Texturing/MatrixTextureTransform.js
@@ -22,7 +22,7 @@ x3dom.registerNodeType(
          * @extends x3dom.nodeTypes.X3DTextureTransformNode
          * @param {Object} [ctx=null] - context object, containing initial settings like namespace
          * @classdesc The MatrixTextureTransform node defines a 2D transformation as a 4x4 matrix that is applied to texture coordinates.
-           This node affects the way textures coordinates are applied to the geometric surface.
+         * This node affects the way textures coordinates are applied to the geometric surface.
          */
         function (ctx) {
             x3dom.nodeTypes.MatrixTextureTransform.superClass.call(this, ctx);
--- a/src/nodes/Texturing/MatrixTextureTransform.js
+++ b/src/nodes/Texturing/MatrixTextureTransform.js
@@ -1,0 +1,57 @@
+/** @namespace x3dom.nodeTypes */
+/*
+ * X3DOM JavaScript Library
+ * http://www.x3dom.org
+ *
+ * (C)2009 Fraunhofer IGD, Darmstadt, Germany
+ * Dual licensed under the MIT and GPL
+ */
+
+/* ### TextureTransform ### */
+x3dom.registerNodeType(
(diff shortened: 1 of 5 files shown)


andreasplesch:
This is tested with both the gltf example at
https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/TextureTransformTest/glTF/TextureTransformTest.gltf

and the x3d example at 
http://www.web3d.org/x3d/content/examples/ConformanceNist/Appearance/TextureTransform/center_primitivesX3dom.xhtml

Here is a build with the changes applied:
https://5cf65c6c3110d6b8e7692a40--x3dom.netlify.com/

tsturm:
Hi @andreasplesch, thanks for this fix!

I'm not sure If I really like this special handling for glTF inside of a X3D-Node.
Maybe we should think about a MatrixTextureTransform-Node which only gets a ready to use matrix, like the MatrixTransform-Node. So that we can move the whole special glTF logic into the glTF loader.

andreasplesch:
I agree, a special node would be nicer. This was just the most convenient with the least changes. Keep in mind, the mode field is completely transparent, x3d would never need to know about it.

My only concern with MatrixTextureTransform would be animation. Animating the rotation, scale, offset of the texture transform can have cool effects, although I am not sure if it is supported right now in gltf. It may be. A MatrixTransform would make it impossible to animate the transform by routing interpolators in x3d.

I guess we could just have a TextureTransformTRS node. It would most closely reflect the glTF extension spec.

Another, secondary consideration is the X3D v4 spec. effort. A new node would be harder to add to the spec. since it needs more language..

I also discovered likely disagreement in how glTF renderers treat the extension:

https://github.com/KhronosGroup/glTF-Sample-Models/issues/225

andreasplesch:
Here is a discussion on animation of matrices in glTF:
https://github.com/KhronosGroup/glTF/issues/1287
The conclusion is that it is not a good idea and not allowed.

andreasplesch:
glTF does not have animation of tex. transforms, for now. So a MatrixTextureTransform node would be ok.  Do you want to give it a try ?

andreasplesch:
Ok, I removed the glTF mode and added instead a MatrixTextureTransform node.  I tested for both glTF and x3d scenes successfully. Take a look.

tsturm:
Looks perfect, thanks!

PR #969, opened 2019-07-16
inline contentType field fixes

andreasplesch:
there were some mixups

### DIFF
--- a/src/nodes/Networking/Inline.js
+++ b/src/nodes/Networking/Inline.js
@@ -171,7 +171,7 @@ x3dom.registerNodeType(
             {
                 for ( var type in this.ContentType )
                 {
-                    if( contentType === this.ContentType[ type ] )
+                    if( type === contentType )
                     {
                         return true;
                     }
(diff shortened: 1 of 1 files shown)


src/nodes/Networking/Inline.js
```diff
@@ -171,7 +171,7 @@ x3dom.registerNodeType(
             {
                 for ( var type in this.ContentType )
                 {
-                    if( contentType === this.ContentType[ type ] )
+                    if( type === contentType )
```
tsturm: if( contentType === this.ContentType[ type ] ) is the correct one here.
Input is "model/x3d+xml", "model/gltf+json", "model/gltf-binary" or "model/x3d+json"
The range of the documentation was wrong at that point

src/nodes/Networking/Inline.js
```diff
@@ -185,7 +185,7 @@ x3dom.registerNodeType(
                 //Return contentType if defined & valid
                 if (this._vf.contentType != "" && this.isValidContentType( this._vf.contentType ) )
                 {
-                    return contentType;
+                    return this.ContentType[this._vf.contentType];
```
tsturm: It should be: return this._vf.contentType;

I already commit the change to master

tsturm:
Already fixed with f65cecf11771fbca46217a218f3461c70eddd699

andreasplesch:
ok

Issue #972, opened 2019-08-04
Background urls out of order


andreasplesch:
The provided urls for back,fron,top,bottom,left,right are used out of order:

https://raw.githack.com/x3dom/x3dom/background_test/doc/misc/iX/sonnensystem_x3dom.html

https://raw.githack.com/x3dom/x3dom/background_test/test/regression-suite/test/cases/cubemap/cubemap.html

elmkni:
Hi Andreas,

I tested the incorrect mapping of the 6 cubemap images myself.

Indeed they are not correctly placed! See my screenshot:

![cubemap_space](https://user-images.githubusercontent.com/53652478/62457322-e1945a80-b77a-11e9-850b-486da1ccec9b.png)

I added my test-files in [cubemap_space.zip](https://github.com/x3dom/x3dom/files/3467112/cubemap_space.zip)

cubemap_space.html
cubemap_space.png
cubemap_space.txt

With best regards,

Elmar

andreasplesch:
Thanks for the confirmation. Fix in #973 .

tsturm:
@andreasplesch Should we add the last commits as a patch to the current stable release?

jamesleesaunders:
Shouldn’t it just follow the semver standard (major, minor, patch) release version numbering? I also feel that having regular patch releases gives a good impression that a prone t is current and active. Maybe not for a single bug but if we collect a few then I think 1.8.1 I think is worth it.
Also the npm version will have to be 1.8.1 as you can not easily update versions on npm.

andreasplesch:
Let's get this out sooner rather than later. I would just agree to what @tsturm wants to do but having a quick 1.8.1 seems cleanest. 1.8.0 could perhaps even be removed from the x3dom.org server if there are constraints.

tsturm:
Okay, then let us collect some bugs and release 1.8.1 in a few weeks.

Issue #971, opened 2019-08-03
Background node cannot handle data-uris for textures (eg. backUrl) anymore!


elmkni:
Background node cannot handle data-uris for textures (eg. backUrl) anymore!

In 1.7.2 everything worked! Could be used to insert textures on the fly with javascript/canvas2D.

 e.g.:

[v3_tt004_hwing.html.txt](https://github.com/x3dom/x3dom/files/3463828/v3_tt004_hwing.html.txt)

![v3_tt004_hwing html](https://user-images.githubusercontent.com/53652478/62412836-f7cdd980-b607-11e9-9032-db96d190d394.png)

andreasplesch:
for experimentation:
https://jsfiddle.net/v5ughdjq/2/

elmkni:
Hi Andreas,

thank you for your prompt answer. I tested already with a complete cubemap:

It works only with square images where width=height (even with scaling="true"),
otherwise webGL throws an error (at least in chrome 76) and only if all 6 images 
are set and not just the backUrl!

But that is not what I want! 

I want to be able to use 1 (one) background image, that statically occupies 
the complete background (streched if nessecary) as was the behavior with 
x3dom until v1.7.2 (which was a not-X3D-compliant feature, but very nice 
and usefull!).

Not a good idea to break a behavior that existed for ~10 years and upon 
which many people rely.

I tested alternatively to set style="background:url(data:....)" in \<X3D\>, which 
works, but makes it impossible to make a x3dom-snapshot including the back-
ground, where the model is rendered into the background image with nice 
antialiasing, which is the reason I used X3DOM (and not X3D) in the first place!

I also tested to set an \<ImageTexture\> on the fly: it works even with non-square
or partially transparent images! Why doesn't it work with \<Background\>?

Is it possible to restore the old behavior in cases where only the backUrl is set 
to a square/non-square image?

By the way: I also noticed, that I cannot copy/paste the debug log any more,
for example to get the viewpoint of a special positure of my model, to insert
it into the html-file.

I would like to use x3dom v1.8.0 because of its new additional features, like
NURBS etc. and because it uses webGL 2.0, but only if the background problem
and the debug-log in-accessibility can be fixed.

With best regards,

Elmar

andreasplesch:
Thanks for the comprehensive report. I suspect the background changes have to do with supporting .dds in the backUrl field for PBR purposes. Unfortunately, I was not involved with these changes. Perhaps @tsturm has a quick fix on hand.

[...]

PR #973, opened 2019-08-04
fix Background face order and undefined eyeIdx

andreasplesch:
addresses #972.
Create cubemap texture in dds order since dds shader is used now.
addresses #971 
guard against undefined eyeIdx

### DIFF
--- a/doc/misc/iX/sonnensystem_x3dom.html
+++ b/doc/misc/iX/sonnensystem_x3dom.html
@@ -14,8 +14,8 @@
 
 
 	<!-- X3DOM -->
-    <script type='text/javascript' src='https://www.x3dom.org/x3dom/release/x3dom.js'></script>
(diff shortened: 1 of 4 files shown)


andreasplesch:
fixed tests:

https://raw.githack.com/x3dom/x3dom/background/test/regression-suite/test/cases/cubemap/cubemap.html

https://raw.githack.com/x3dom/x3dom/background/doc/misc/iX/sonnensystem_x3dom.html

tsturm:
Hi @andreasplesch,

sorry for the late response. I was not in the office the last days.
I have merged your pull request but I will change it a little bit afterwards.

The check for eyeIdx equals undefined is one solution, but I think its better to update the BackgroundShader so it also supports VR.

andreasplesch:
Yes, a proper fix for all BackgroundShaders would be best. Ah, I see you already started the fixes.

Issue #227, opened 2013-10-25
Leak when changing DOM tree


stepasite:
Hi,

when removing nodes from DOM tree, they seem to remain in detached DOM tree and thus they never get freed by gc from memory.

Problem can be seen e.g. in:

http://x3dom.org/x3dom/example/x3dom_addRemoveNode.html

example, but it is not this example specific, it is general.

Observation in Firefox:

1) open x3dom_addRemoveNode.html example in browser
2) there are 3 #nodes (Firefox takes about 90 MB in memory)
3) Add over 2000 #nodes (Firefox takes about 190 MB in memory)
4) Remove all nodes (Firefox still occupies about 200 MB)

The problem is not Firefox specific, but it happens also in Chrome. When profiling in Chrome, many arrays remain in memory as well as many nodes in detached DOM tree.

This is not problem for small scene like this, but become more serious when replacing complex scenes containing big amount of data.

I tried to find where the references to nodes are kept, but without success. NodeNamespace does not seem to be the cause.

Thanks in advance for any help.

StinsonZhao:
It's 2018! But the problem is still here.

andreasplesch:
The example above is using 1.7.1 . Likely, the problem persists with current master as well but it would be very constructive to confirm that. Could you do that ?

tomssd:
Has there been any progress on this issue?  I am trying to load and unload a lot of inlined nodes.

tomssd:
Setting the load attribute to false of the inline node does appear to successfully remove the object from the scene graph.  I have tried removing the element from the DOM tree and destroying any references.  The memory utilization drops slightly but the net result from adding/removing the models is an increase in memory until the browser eventually crashes.

tomssd:
In Version 1.7.3:

tomssd:
I have been testing 1.8.1

THIS WORKS:
http://cowanland.com/memtest/memory_test_nodes.html

Adding and removing basic nodes from the DOM is no longer a problem.  The memory does clear out after the node has been removed.

THIS DOES NOT WORK:
http://cowanland.com/memtest/memory_test_inlines.html

Removing inline nodes from the DOM is a problem.  
The memory is not cleared out.  

I have tried several things without success (recursively removing all child nodes, setting the inline load = false, setting the inline render = false).

andreasplesch:
I am tracking down memory leaks with the Chrome devtools, in a custom branch, which is quite something.

Here is a current build of this custom branch:
[...]

PR #1007, opened 2019-10-21
lgtm warnings

andreasplesch:
solves #227


andreasplesch:
A second look would be good since the required changes are more significant.

andreasplesch:
Thanks for the quick review, and apologies for the misleading title from the other branch. I changed the title retroactively, for the record.

PR #1011, opened 2019-10-27
Text tuning

andreasplesch:


jamesleesaunders:
@andreasplesch
I think this may have broken text labels?

See: https://raw.githack.com/jamesleesaunders/d3-x3d/master/examples/X3DOM/chart/AreaChartMultiSeries.html

jamesleesaunders:
On Safari iPhone

andreasplesch:
I do not have an iphone to test with. Is it just Safari ?

Screenshot ?

Maybe the mipmap minfilter ?

andreasplesch:
https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/generateMipmap

says safari ios and edge do not support that on webgl2 (but they do on webgl1 ?)

tsturm:
WebGL 2.0 should be disabled on ios devices. Looks good on Windows (Chrome/Firefox) and Android(Chrome/Firefox)

jamesleesaunders:
IMG_2060

tsturm:
This build looks good in IE11.
Which shows the black boxes with the current master, too.

@jamesleesaunders In IE11 there is some error inside your d3-x3d.js

Should we scratch IE support in the future?

andreasplesch:
Yes, IE11 support could be removed, Edge hopefully will not require special treatment.

andreasplesch:
Looks good to me with pot sizes on browserstack (iphone 6s, safari):

image

jamesleesaunders:
Cheers guys,
Not had chance to check myself yet but looks good.
IE11 support... yes agree could potentially drop support, especially if it is holding things back.

andreasplesch:
It would be good to check on a newer iphone/safari: https://raw.githack.com/andreasplesch/d3-x3dom/51d6451e4728131215b0b0ccd7333341f3816423/examples/X3DOM/chart/AreaChartMultiSeries.html uses the new build.

jamesleesaunders:
Just tested on iPhone XS IOS 13.1 with your build @andreasplesch and it all looks good now :-) Thanks.

andreasplesch:
ok, I will prepare a PR. @tsturm : do you have an opinion if nextbestpowerof2 which selects down if the lower pot is closer,or nexthigherpowerof2 which always selects up should be preferred ?

andreasplesch:
Testing on Android shows navigation on mobile is now awkward because it moves both the camera and the web page. I think this preventDefault related ? If so, is there a way to get it back ? (zoom wheel is also awkward).

tsturm:
You can use the x3dom.Utils.scaleImage function which uses the nexthigherpowerof2 internally.

I will check the preventDefault issue tomorrow.

andreasplesch:
thanks. Thinking about it, one can always use a higher quality setting, so I think sometimes downsampling is fine. Also, the text rendering already uses a canvas2d, so the scaling is easy there.

PR #1018, opened 2019-11-08
lgtm fixes

andreasplesch:


src/nodes/NURBS/x3dom-nurbs-web-worker.js
```diff
@@ -28,11 +28,11 @@
         {
             if ( e.data[ 11 ] )
```
jamesleesaunders: Variable scoping, strictly speaking this variable should be declaired outside the if block.
```
var basisFunsCache;
if (..) {
   basisFunsCache = e.data[ 11 ];
} else {
   basisFunsCache = new Map();
}
```
Not so much an issue with car, but would become a problem if/when move to ES6 let?

andreasplesch: Yes, I just did not want to introduce another line. Moving to let will be problematic in many places.

PR #1058, opened 2020-04-16
Fix bug on unload function

rsm-gh:
Correct the error `x3dom.canvases[i].doc is null`. It was discoveted with Firefox 75.0 x64 For Arch Linux 1.0 when the message "X3DOM is not supported" appears.

### DIFF
--- a/src/Main.js
+++ b/src/Main.js
@@ -233,7 +233,14 @@ x3dom.userAgentFeature = {
         {
             for ( var i = 0; i < x3dom.canvases.length; i++ )
             {
-                x3dom.canvases[ i ].doc.shutdown( x3dom.canvases[ i ].gl );
+                
+                var canvase = x3dom.canvases[ i ];
+            	
+            	if (canvase.doc != null)
+                {
+            		canvase.doc.shutdown(canvase.gl);
+            	}
+                
             }
             x3dom.canvases = [];
         }


andreasplesch:
Thanks. Could you please test Chrome or Chromium as well ?

rsm-gh:
@andreasplesch with chromium I'm not able to reproduce the bug because X3DOM is being correctly loaded. Actually I need to find out why it is not working anymore with Firefox.

andreasplesch:
Do you have an example ?

rsm-gh:
@andreasplesch, what do you mean by an example?  The software that I develop can be found here: [https://senties-martinelli.com/software/cad-viewer]()

But it depends of the client web browser. The most that I can give you is an screenshoot

rsm-gh:
Screenshot_2020-04-16_20-45-41
Screenshot_2020-04-16_20-47-40

andreasplesch:
Does it work with FF 74.0 ? It works for me on linux mint with 74 and on debian with 68 esr.

FF 75.0 also works for me on Windows.

How does the null error affect your app ?

rsm-gh:
@andreasplesch you can implement the fix as you want, I'm just reporting it. When the bug happens, since it makes an JS error all the JS of the website it is not any more executed which is kinda problematic.

Firefox was working alright until 4-5 days, then Webgl seems disabled. I still need to check this, I ignore it it is because of a version change or a patch.

andreasplesch:
https://github.com/x3dom/x3dom/commit/6b69d0257583c8cab670e6f6eb864b9465c728ca applies the suggested fix.

Issue #1059, opened 2020-05-17
Incorrect tessellation of rational NURBS curves and surfaces


cmasia:
## Summary

Rational NURBS curves  and surfaces are not tessellated correctly. NURBS are rational when (at least) one of the weights is different from the default value of 1.

## Steps to reproduce

Problem is illustrated for curves (NurbsCurve node) in this [X3DOM 1.8.1 NURBS Curve evaluation bug](https://codepen.io/cmasia/pen/XWmxKLo) codepen.

This [X3DOM 1.8.1 NURBS Surface evaluation bug](https://codepen.io/cmasia/pen/JjYmNWJ) codepen replicates the problem for surfaces (NurbsPatchSurface node).

## Expected result

In the curve codepen, the red curve should look like the green one (which is an Arc2D node). Both are circle arcs, although on translated planes.

For the surface codepen, the geometry should be a arc of a cylinder (in this example, it is the extruded version of the arc in the first example).

## Actual result

The NURBS arc (red) does not look like an arc:
![image](https://user-images.githubusercontent.com/26604712/82141039-83c98e80-9833-11ea-95a9-6e3aa60921d7.png)

The NURBS cylinder arc does not look like a cylinder arc:
![image](https://user-images.githubusercontent.com/26604712/82141060-a0fe5d00-9833-11ea-81d4-8f1cfd22a32b.png)

## Additional information

The evaluation of points on the NURBS seems to be the culprit, but only for rational NURBS entities. Polynomial NURBS (all weights equal to 1.0) are rendered correctly.

The equation for a NURBS curve is:
![image](https://user-images.githubusercontent.com/26604712/82141188-72cd4d00-9834-11ea-8e8c-aeda69d54e69.png)

For a NURBS surface, the equation is:
![image](https://user-images.githubusercontent.com/26604712/82141216-a0b29180-9834-11ea-89ad-ca703238f767.png)

## Analysis and fix

After locating the code that evaluates a point on NURBS curves and surfaces, it appears that the weights (w_i) in the numerator of the formulas are not applied.

In x3dom/src/nodes/NURBS/NurbsCurve.js, function curvePoint3DH(). 

```javascript
                for ( k = 0; k <= p; k++ )
                {
                    i = indu + k;
                    temp[ 0 ] += Nu[ k ] * P[ i ].x;
                    temp[ 1 ] += Nu[ k ] * P[ i ].y;
                    temp[ 2 ] += Nu[ k ] * P[ i ].z;
[...]

PR #1060, opened 2020-05-17
fix for #1059: incorrect tessellation of rational NURBS

cmasia:
Weights where not taken into account when evaluating points of NURBS curves
and surfaces, leading to incorrect tessellations when one or more weights
were different from the default value of 1.0.

Problem is solved by multiplying the control points coordinates by the
weights in the numerator part of the formula.


andreasplesch:
looks good.

PR #1070, opened 2020-06-05
Remove new Date().getTime() fallback in start-/stopMeasure

campersau:
as performance.now is always available since https://github.com/x3dom/x3dom/commit/9e1f483d3f8e3e39385a91933d0fec4935d82cef

https://github.com/x3dom/x3dom/blob/9edcd6e2f6fb93bfddad94981131b144242b8889/src/util/Utils.js#L22-L34


andreasplesch:
Looks good.

Just for future reference

https://caniuse.com/#search=performance.now

PR #1072, opened 2020-06-11
If someone can tell me what to do with these beside do a pull request, I would be happy!

coderextreme:


andreasplesch:
There seem to be no changes compared to the master. There is nothing to do.

andreasplesch:
Please reopen as neessary.

PR #1077, opened 2020-06-24
Replace 1.5707963 with Math.PI

FrankBuchholz:
I'm always confused when I see 'magic' numbers in code. 
What is the meaning of `1.57`?
Does this represent 90° or approximately 90° (because high precision does not matter) or a little bit less than 90° (which might be important to avoid singularity issues)?

In the code we mostly see `Math.PI / 2` instead and sometimes `Math.PI / 2.0` (which looks better as it's a float anyway). 

Therefore I've replaced the remaining different constants `1.57`, `1.570796`, `1.5707963`, `1.57079633` by `Math.PI / 2.0` 

(Exception: in ShaderDynamic.js the constants are required for the gl command string.)

Limited test: My program is still working and it's using some of these pieces of code. However, I've no idea how to produce test cases for all changes.

### DIFF
--- a/src/util/VRControllerManager.js
+++ b/src/util/VRControllerManager.js
@@ -128,7 +128,7 @@ x3dom.VRControllerManager.prototype.fit = function ( viewarea, vrDisplay )
 
     var aspect =  Math.min( viewarea._width / viewarea._height, 1 );
 
-    var tanfov2 = Math.tan( 1.57 / 2.0 );
+    var tanfov2 = Math.tan( Math.PI/4.0 );
     var dist = bsr / tanfov2 / aspect;
 
     viewarea._movement = viewDir.multiply( -dist );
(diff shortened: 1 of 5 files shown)


src/util/VRControllerManager.js
```diff
@@ -128,7 +128,7 @@ x3dom.VRControllerManager.prototype.fit = function ( viewarea, vrDisplay )
 
     var aspect =  Math.min( viewarea._width / viewarea._height, 1 );
 
-    var tanfov2 = Math.tan( 1.57 / 2.0 );
+    var tanfov2 = Math.tan( Math.PI / 4.0 );
```
andreasplesch, line 131: I believe tanfov2 means tangents of field of view divided by 2.
`Math.tan( 0.5 * Math.PI / 2.0` would be closer to that.

andreasplesch:
I believe 1.57 is used just for brevity. Math.PI makes more sense.

PR #1079, opened 2020-06-27
Extend log messages

FrankBuchholz:
Concerning pick mode:
"color" shows the RGB colors.
"idbuf", "idbuf24", "box" show the position.
"idbufid", "texcoord" don't show the position, therefore I decided to name it simply 'data'.

Concerning Background node:
It is not useful to show the sum count of explicit plus implicit colors or angels - the user would not recognize and understand these numbers. However, it might help to explain how to correct the data.
This condition has to be true:
(skyAngle.lenght == skyColor.length - 1) && (groundAngle.length == groundColor.length -1)

### DIFF
--- a/src/gfx_webgl.js
+++ b/src/gfx_webgl.js
@@ -1049,9 +1049,9 @@ x3dom.gfx_webgl = ( function ()
 
                 if ( sky.length != colors.length )
                 {
-                    x3dom.debug.logError( "Number of background colors (" + colors.length
-                        + ") and corresponding angles (" + sky.length
-                        + ") are different!" );
+					// 
+                    x3dom.debug.logError( "Number of background colors and corresponding angles do not match.\n"
+						+ "You have to define one angle less than the count of RGB colors because the angle 0° is added automatically." );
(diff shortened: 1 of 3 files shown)


src/gfx_webgl.js
```diff
@@ -1049,7 +1049,8 @@ x3dom.gfx_webgl = ( function ()
 
                 if ( sky.length != colors.length )
                 {
-                    x3dom.debug.logError( "Number of background colors and corresponding angles are different!" );
+                    x3dom.debug.logError( "Number of background colors and corresponding angles do not match.\n"
+                        + "You have to define one angle less than the count of RGB colors because the angle 0° is added automatically." );
```
andreasplesch, line 1053: That's ok. Let's just keep logging concise. One can assume that users are familiar with the spec. in general.

FrankBuchholz: You are right, the spec tells everything in great detail: 
https://www.web3d.org/documents/specifications/19775-1/V3.3/Part01/components/enveffects.html#Backgrounds

Issue #777, opened 2017-10-18
handling Viewpoint PgDn/PgUp/Home/End keys


brutzman:
Here is a fun model that includes a set of ordered viewpoints:

http://x3dgraphics.com/examples/X3dForAdvancedModeling/ShayDPixel/ShayDPixelBobbleHeadBoxSiggraph2017Index.html

Of specific interest is X3DOM version:

http://x3dgraphics.com/examples/X3dForAdvancedModeling/ShayDPixel/ShayDPixelBobbleHeadBoxSiggraph2017X3dom.xhtml

1.  If you use PgDn/PgUp on the initial scene to go to next/previous Viewpoint, they work.  However, the HTML page also scrolls, which in turn moves the X3DOM space out of sight as the whole HTML page scrolls.  Confusing and disorienting, unfortunately.

Recommendation, if possible: when mouse isOver the X3DOM window, trap PgDn/PgUp from also getting passed to HTML browser.

2.  Select "Zoom" button in upper left.  Now again try PgDn/PgUp, which should go in order front/left side/back/right side/bottom/top then repeat.

X3DOM response to the PgDn/PgUp keys is reversed, and should be fixed to go next/previous Viewpoint respectively.

3. Home key (initial viewpoint) and End key (final viewpoint) do not seem to be supported yet.  Recommend adding.

4. Viewpoints are the #1 user navigation method in X3D scenes, so fixing these problems will be great.

References:

Annex G, Recommended navigation behaviours, G.2 Select from multiple viewpoints
www.web3d.org/documents/specifications/19775-1/V3.3/Part01/behaviours.html#SelectFromMulitpleViewpoints

> This annex recommends using the following keys:
>
> HOME    Initial Viewpoint PGDN    Next Viewpoint PGUP    Previous Viewpoint END     Final Viewpoint 

23 Navigation component, 23.2 Concepts
http://www.web3d.org/documents/specifications/19775-1/V3.3/Part01/components/navigation.html#Concepts

X3D Tooltips: Viewpoint
http://www.web3d.org/x3d/content/X3dTooltips.html#Viewpoint

Having fun with X3DOM!  8)

andreasplesch:
Let's break this up:

[ ] PgUp/PgDn reversed

...

andreasplesch:
Found some time and could fix these navigation issues. Should propagate to x3dom.org in 24h.

PR #1081, opened 2020-06-30
Extend help shown in debug window

FrankBuchholz:
Add missing key commands to help info shown in debug window.

### DIFF
--- a/src/X3DDocument.js
+++ b/src/X3DDocument.js
@@ -399,11 +399,11 @@ x3dom.X3DDocument.prototype.onKeyPress = function ( charCode )
             {
                 states.display();
             }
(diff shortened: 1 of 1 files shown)


andreasplesch:
home and end do not work that well because the browser also scrolls the page. Do you think these keys should still be listed ?

FrankBuchholz:
I use a full screen X3DOM scene and have therefore no issues with scrolling... I guess this is the reason why I didn't realized potential conflicts. Anyway, the keys are active and you can use them to swap viewpoints.

I just found an old issue about these keys:
https://github.com/x3dom/x3dom/issues/777

I follow your advice.

andreasplesch:
Did you want to make a change to this PR ?

andreasplesch:
Let's keep it as is.

PR #1082, opened 2020-06-30
Correct axis definition "negY"

FrankBuchholz:
Axis  definition "negY" evaluates to ( 1, -1,  0 ) instead of ( 0, -1,  0 ).

In addition we could think about accepting axis definitions directly instead of symbolic names like "negY": In the default case of the switch statement we could check if parameter axis has type `SFVec3f` or can be resolved to this type. Well, that would be an extension of the API which requires corresponding documentation. And in this case we should do the same in `x3dom.Viewarea.prototype.showAll`.


andreasplesch:
I tested the change with the runtime.html example which uses showObject. Looks good.

Generally, it tends to be better to only allow well defined values for parameters to avoid a lot of issues. It is true that the fitting functions are really useful and probably work with a general axis direction which would be a relatively easy and valuable addition.

PR #1091, opened 2020-07-14
Small documentation updates

FrankBuchholz:
While analyzing x3dom, i.e. about defining background cube pictures and about copying ideas from the helicopter mode for my own navigation mode, I made some comments into the documentation which might be useful for the master, too.

- The background cube is defined in x3d, it's not added by x3dom. 
see 
https://www.web3d.org/documents/specifications/19775-1/V3.3/Part01/components/enveffects.html#Background
- I didn't found a way to use the `scaling` attribute of the `Background` note but a least I can state, that's a boolean attribute.
see `gfx_webgl.js`
`if ( bgnd._vf.scaling && bgnd._webgl.texture.ready )` ...

- The array attribute `typeParams`  of the `NavigationInfo` note has up to 7 values.

side remarks:
The scaling of the background cube pictures seemed to have changed between versions but I didn't found the reason yet (otherwise I would try to offer a PR for it): With the current version of x3dom you see a heavily zoomed-in picture which looks quite bad. It's difficult to demonstrate it in my program but you can see the same effect here: http://antorof.github.io/SolarSystem-X3Dom/ 
The background should look like this:
http://antorof.github.io/SolarSystem-X3Dom/sistemasolar/resources/space_BK.jpg
http://antorof.github.io/SolarSystem-X3Dom/sistemasolar/resources/space_LF.jpg
etc.

### DIFF
--- a/src/fields.js
+++ b/src/fields.js
@@ -420,7 +420,7 @@ x3dom.fields.SFMatrix4f.prototype.setRotate = function ( quat )
 };
 
 /**
- * Creates a new matrix from a column major string representation, with values separated by commas
+ * Creates a new matrix from a column major string representation, with values separated by commas or whitespace
  *
  * @param {String} str - string to parse
  * @returns {x3dom.fields.SFMatrix4f} the new matrix
@@ -1690,7 +1690,7 @@ x3dom.fields.SFMatrix4f.prototype.toString = function ()
 
 /**
  * Fills the values of this matrix from a string, where the entries are separated
- * by commas and given in column-major order.
+ * by commas or whitespaces and given in column-major order.
  *
  * @param {String} str - the string representation
  */
--- a/src/fields.js
+++ b/src/fields.js
@@ -1690,7 +1690,7 @@ x3dom.fields.SFMatrix4f.prototype.toString = function ()
 
 /**
  * Fills the values of this matrix from a string, where the entries are separated
- * by commas or whitespaces and given in column-major order.
+ * by commas or whitespace, and given in column-major order.
  *
  * @param {String} str - the string representation
  */
--- a/src/nodes/EnvironmentalEffects/Background.js
+++ b/src/nodes/EnvironmentalEffects/Background.js
@@ -30,7 +30,7 @@ x3dom.registerNodeType(
             x3dom.nodeTypes.Background.superClass.call( this, ctx );
 
             /**
-             * Image shown at the back of the background cube 
+             * Image shown at the back of the background cube
              * @var {x3dom.fields.MFString} backUrl
              * @memberof x3dom.nodeTypes.Background
              * @initvalue []
@@ -41,7 +41,7 @@ x3dom.registerNodeType(
             this.addField_MFString( ctx, "backUrl", [] );
 
             /**
-             * Image shown at the bottom of the background cube 
+             * Image shown at the bottom of the background cube
              * @var {x3dom.fields.MFString} bottomUrl
              * @memberof x3dom.nodeTypes.Background
              * @initvalue []
@@ -52,7 +52,7 @@ x3dom.registerNodeType(
             this.addField_MFString( ctx, "bottomUrl", [] );
 
             /**
-             * Image shown at the front of the background cube 
+             * Image shown at the front of the background cube
              * @var {x3dom.fields.MFString} frontUrl
              * @memberof x3dom.nodeTypes.Background
              * @initvalue []
@@ -63,7 +63,7 @@ x3dom.registerNodeType(
             this.addField_MFString( ctx, "frontUrl", [] );
 
             /**
-             * Image shown at the left of the background cube 
+             * Image shown at the left of the background cube
              * @var {x3dom.fields.MFString} leftUrl
              * @memberof x3dom.nodeTypes.Background
              * @initvalue []
@@ -74,7 +74,7 @@ x3dom.registerNodeType(
             this.addField_MFString( ctx, "leftUrl", [] );
 
             /**
-             * Image shown at the right of the background cube 
+             * Image shown at the right of the background cube
              * @var {x3dom.fields.MFString} rightUrl
              * @memberof x3dom.nodeTypes.Background
              * @initvalue []
@@ -85,7 +85,7 @@ x3dom.registerNodeType(
             this.addField_MFString( ctx, "rightUrl", [] );
 
             /**
-             * Image shown at the top of the background cube 
+             * Image shown at the top of the background cube
              * @var {x3dom.fields.MFString} topUrl
              * @memberof x3dom.nodeTypes.Background
              * @initvalue []
--- a/src/nodes/EnvironmentalEffects/Background.js
+++ b/src/nodes/EnvironmentalEffects/Background.js
@@ -96,7 +96,7 @@ x3dom.registerNodeType(
             this.addField_MFString( ctx, "topUrl", [] );
 
             /**
-             * Scale images of the background cube (### verification needed as I didn't found any example running trought the corresponding coding)
+             * Scale images of the background cube
              * @var {x3dom.fields.MFString} scaling
              * @memberof x3dom.nodeTypes.Background
              * @initvalue false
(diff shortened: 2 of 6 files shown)


src/nodes/EnvironmentalEffects/X3DBackgroundNode.js
```diff
@@ -41,6 +41,8 @@ x3dom.registerNodeType(
 
             /**
              * Color of the ground
+             * Multiple color values are possible.
+             * You have to define 1 color more that you define ground angles.
```
andreasplesch, line 45: "Define 1 more color than the number of ground angles."

src/nodes/Navigation/NavigationInfo.js
```diff
@@ -62,6 +62,14 @@ x3dom.registerNodeType(
 
             /**
              * Specifies the view angle and height for helicopter mode and min/max rotation angle for turntable in ]0, PI[, starting from +y (0) down to -y (PI)
+             * [0] helicopter vertical view angle theta
+             * [1] helicopter height
+             * [2] min vertical view angle theta
```
andreasplesch, line 67: turntable theta max

src/nodes/Navigation/NavigationInfo.js
```diff
@@ -62,6 +62,14 @@ x3dom.registerNodeType(
 
             /**
              * Specifies the view angle and height for helicopter mode and min/max rotation angle for turntable in ]0, PI[, starting from +y (0) down to -y (PI)
+             * [0] helicopter vertical view angle theta
+             * [1] helicopter height
+             * [2] min vertical view angle theta
+             * [3] max vertical view angle theta
```
andreasplesch, line 68: turntable theta min

src/nodes/Navigation/NavigationInfo.js
```diff
@@ -62,6 +62,14 @@ x3dom.registerNodeType(
 
             /**
              * Specifies the view angle and height for helicopter mode and min/max rotation angle for turntable in ]0, PI[, starting from +y (0) down to -y (PI)
+             * [0] helicopter vertical view angle theta
+             * [1] helicopter height
+             * [2] min vertical view angle theta
+             * [3] max vertical view angle theta
+             * [4] turntable min orbit rotation angle phi (yaw)  (### needs verification)
```
andreasplesch, line 69: turntable phi max

src/nodes/EnvironmentalEffects/X3DBackgroundNode.js
FrankBuchholz: English is so simple (but difficult for a German) ;-)

src/nodes/Navigation/NavigationInfo.js
FrankBuchholz: I guess `typeParams[ 2 ]` is the min value and `typeParams[ 3 ]` is the max value:

`NavigationInfo.js`
```
getTypeParams : function ()
            {
                var length = this._vf.typeParams.length;

                var theta  = ( length >= 1 ) ? this._vf.typeParams[ 0 ] : -0.4;
                var height = ( length >= 2 ) ? this._vf.typeParams[ 1 ] : 60.0;
                var minAngle = ( length >= 3 ) ? this._vf.typeParams[ 2 ] : x3dom.fields.Eps;
                var maxAngle = ( length >= 4 ) ? this._vf.typeParams[ 3 ] : Math.PI - x3dom.fields.Eps;
```

This match to the usage here:
`TurntableNavigation.js`
```
    // clamp theta
    theta = Math.max( navi._vf.typeParams[ 2 ], Math.min( navi._vf.typeParams[ 3 ], theta ) );
```

What do you think about:
```
             * [2] turntable min vertical view angle theta
             * [3] turntable max vertical view angle theta
```

src/nodes/Navigation/NavigationInfo.js
andreasplesch, line 69: [4] and [5] is also general zoom min/max  if all 7 params are provided: https://github.com/x3dom/x3dom/blob/1f251fd37c760f3aac892cc2398ac8cec90c54c3/src/nodes/Navigation/modes/DefaultNavigation.js#L39

FrankBuchholz: This part looks weird in function `x3dom.TurntableNavigation.prototype.calcOrbit`:
- there is some irrelevant code
- typeParams[ 4/5 ] can be used either used as min/max or as max/min

```
x3dom.TurntableNavigation.prototype.calcOrbit = function ( view, alpha, beta, absolute )
{
    var navi = this.navi;

    view._up   = view._flyMat.e1();
    view._from = view._flyMat.e3();

    var offset = view._from.subtract( view._at );

    var phi,
        theta;

    if ( absolute == true )
    {
        phi = beta;
        theta = alpha;
    }
    else
    {
        // angle in xz-plane
        // irrelevant because phi get's replaced before usage
        phi = Math.atan2( offset.x, offset.z );

        // angle from y-axis
        // irrelevant because theta get's replaced before usage
        theta = Math.atan2( Math.sqrt( offset.x * offset.x + offset.z * offset.z ), offset.y );

        phi -= beta;
        theta -= alpha;
    }

    // clamp theta
    // minTheta = typeParams[ 2 ]; maxTheta = typeParams[ 3 ]
    theta = Math.max( navi._vf.typeParams[ 2 ], Math.min( navi._vf.typeParams[ 3 ], theta ) );

    if ( navi._vf.typeParams[ 4 ] <= navi._vf.typeParams[ 5 ] {
        // minPhi = params[4], maxPhi = param[5]
        phi = Math.max( navi._vf.typeParams[ 4 ], Math.min( navi._vf.typeParams[ 5 ], phi ) );
    }
    else
    {
        // maxPhi = params[4], minPhi = param[5]
        if      ( beta > 0 && phi < navi._vf.typeParams[ 4 ] && phi > navi._vf.typeParams[ 5 ] ) {
            // Let's take the greater value if phi is within the boundaries
            phi = navi._vf.typeParams[ 4 ];
         }
        else if ( beta < 0 && phi < navi._vf.typeParams[ 4 ] && phi > navi._vf.typeParams[ 5 ] ) {
            // Let's take the smaller value if phi is within the boundaries
            phi = navi._vf.typeParams[ 5 ];
        }
    }
[...]
```

In any case, I guess it would be better to replace these positional parameters with named attributes...

Until then: I guess concerning the documentation it's better to omit these details, therefore I take them out of this PR.

src/nodes/EnvironmentalEffects/Background.js
```diff
@@ -41,65 +41,65 @@ x3dom.registerNodeType(
...
              * @range [URI]
-             * @field x3dom
+             * @field x3d
              * @instance
              */
             this.addField_MFString( ctx, "topUrl", [] );
 
             /**
-             *
+             * Scale images of the background cube (### verification needed as I didn't found any example running trought the corresponding coding)
```
FrankBuchholz, line 99: I still do not figured out in current version of x3dom how to scale (or not to scale) the background image cube in a way that these pics are not zoomed-in.

andreasplesch: The zoom is unrelated to scale. The zooming comes from a change in the size of the sphere used as the surface for the background texture, from 10000 to 1000. I think 1000 worked better for VR. For more control it is preferable to just use a big sphere in the scene instead of background.

andreasplesch:
getPagePos is now available, as an alias for getCanvasPos.

src/nodes/EnvironmentalEffects/Background.js
FrankBuchholz: I could imagine that it would be better to see this constant 1000 at the caller within `Context.prototype.setupScene`.
Maybe it even would be better to add an optional `radius` attribute to node `Background`.

andreasplesch: Yes, I think that this is a good idea. The current setup is really obscure.

src/fields.js
```diff
@@ -1690,7 +1690,7 @@ x3dom.fields.SFMatrix4f.prototype.toString = function ()
 
 /**
  * Fills the values of this matrix from a string, where the entries are separated
- * by commas and given in column-major order.
+ * by commas or whitespaces and given in column-major order.
```
andreasplesch, line 1693: 'whitespace' is also the plural.

PR #1098, opened 2020-07-22
Background Cube

FrankBuchholz:
Background Cube images look different between x3dom 1.7.2 (good) and  1.8.1 (ugly). The reason was not, as initially suspected, the size of the generated background sphere - this size is in fact irrelevant- but because the wrong shader name `BACKGROUND_CUBETEXTURE_DDS` which is used even if you do not provide an dds file.

After replacing the shader name, the images look as expected and all images are connected correctly to fill the cube.
However, there is still one more difference: the view is flipped upside down.

So far I tested only using https://github.com/fblupi/x3d-solar-system with the original images and another background picture set.

The current commit contains some additional comments which I will remove later.

### DIFF
--- a/src/gfx_webgl.js
+++ b/src/gfx_webgl.js
@@ -936,17 +936,22 @@ x3dom.gfx_webgl = ( function ()
 
                 bgnd._webgl.primType = gl.TRIANGLES;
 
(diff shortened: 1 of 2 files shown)


andreasplesch:
see also #973

PR #1101, opened 2020-07-24
Added documentation, introduced SFRotation class

KY-ARVR:
Documentation was added, the SFRotation class introduced, the SFImage class amended, and a few unimplemented methods implemented.

### DIFF
--- a/src/fields.js
+++ b/src/fields.js
@@ -789,8 +789,8 @@ x3dom.fields.SFMatrix4f.fromArray = function ( a )
 {
     var m = new x3dom.fields.SFMatrix4f();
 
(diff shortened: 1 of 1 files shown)


andreasplesch:
Thanks, looks good. Do you have a need for the SFRotation type ? Usually, it is more convenient to just use Quaternions or Matrices. x3dom converts SFRotation field values immediately to Quaternions.

Did you test the SFRotaton methods ? It may be possible to avoid additional math by going through existing Quaternion or other methods.

X3D SAI has a few more constructors and methods:

https://www.web3d.org/documents/specifications/19777-1/V3.3/Part1/functions.html#SFRotationInstanceCreationFunction
and below

Could you please split this up, so it is easier to go through ?
Perhaps one PR for documentation changes, one PR for SFRotation methods and one PR for everything else ?

Please use
```
npm run lint
npm run lint-fix
```

to fix the formatting.

PR #1107, opened 2020-08-07
Corrections of documentation, introduction of new functions

KY-ARVR:
Corrections of the documentation:
  - Amended four occurrences of ``@params'' to ``@param''.
  - Amended three occurrences of ``@return'' to ``@returns''.

Introduction of new functions:
  - Added the function ``x3dom.fields.SFMatrix4f.prototype.setAt''.
  - Added the function ``x3dom.fields.SFColor.prototype.toUint''.
  - Added the function ``x3dom.fields.SFColor.prototype.setFromUint''.
  - Added the function ``x3dom.fields.SFColorRGBA.prototype.setFromUint''.


src/fields.js
```diff
@@ -988,6 +988,21 @@ x3dom.fields.SFMatrix4f.prototype.at = function ( i, j )
...
+/**
+ * Sets the value of this matrix at a given position.
+ *
+ * @param {Number} i - row index (starting with 0)
+ * @param {Number} j - column index (starting with 0)
+ * @param {Number} newEntry - the new value to store at position (i, j)
+ * @returns {x3dom.fields.SFMatrix4f} this modified matrix
+ */
+x3dom.fields.SFMatrix4f.prototype.setAt = function ( i, j, newEntry )
+{
```
andreasplesch, line 1000: Do we need this method somewhere in the code base ? With js, performance in the render loop also needs to be considered.

KY-ARVR: The method ``x3dom.fields.SFMatrix4f.prototype.setAt'' is not mandatorily necessary; it merely acts in two capacities: It provides symmetry as the setter variant of the already extant getter ``x3dom.fields.SFMatrix4f.prototype.at'' method, and it might be employed in the method ``x3dom.fields.SFMatrix4f.prototype.polarDecompose'', which in its last loop retrieves and sets matrix elements by its indices. The additional methods for ``SFColor'' and ``SFColorRGBA'' are intended as more efficient alternatives to color modifications in circumstances where no color instance exists to retrieve the components from, thus resorting to hopefully more efficient number (integer) objects. Yet, these new methods can be omitted, too.

andreasplesch: ok, makes sense. Could you revise the last loop in polarDecompose to use the new method, please ? polarDecompose seems to be used in .getTransform which is used quite a bit. So we will need to test example scenes.

The Uint color methods would actually use js Number objects which are double floats. But these would still be more efficient than the rgb objects, and one can instance a color with hex. notation: .setFromUint(0xaabbcc) or with parseInt which seems nice.

src/fields.js
```diff
@@ -1450,11 +1465,11 @@ x3dom.fields.SFMatrix4f.prototype.spectralDecompose = function ( SO, k )
...
+                    var a = SO.at( j, p );
+                    var b = SO.at( j, q );
 
-                    SO[ "_" + j + p ] -= s * ( b + tau * a );
-                    SO[ "_" + j + q ] += s * ( a - tau * b );
+                    SO.setAt( j, p, SO.at( j, p ) - s * ( b + tau * a ) );
+                    SO.setAt( j, q, SO.at( j, q ) + s * ( a - tau * b ) );
                 }
             }
         }
```
andreasplesch, line 1475: Ok. This could be perhaps a bit slower but hopefully not noticeable. These functions are used for default navigation, eg. every frame. Please test just a view example scenes with these changes.

andreasplesch: Did  you have a chance to these changes ?

andreasplesch: oh well, let's just go for it then.

PR #1141, opened 2020-11-20
Added HAnim containerFields

coderextreme:
Added containerFields for USE sites, segment and joints

### DIFF
--- a/src/util/json/JSONParser.js
+++ b/src/util/json/JSONParser.js
@@ -300,12 +300,12 @@ x3dom.JSONParser.prototype.ConvertToX3DOM = function ( object, parentkey, elemen
             else
             {
                 // this.ConvertObject( key, object, element );
-		if (key.indexOf("HAnim") === 0 && key !== "HAnimHumanoid" && typeof object[key]['@USE'] != 'undefined') {
-			object[key]['@containerField'] = key.substr(5).toLowerCase()+"s";
-			this.ConvertObject(key, object, element, object[key]['@containerField']);
-		} else {
-			this.ConvertObject(key, object, element);
-		}
+                if (key.indexOf("HAnim") === 0 && key !== "HAnimHumanoid" && typeof object[key]['@USE'] != 'undefined') {
+                    object[key]['@containerField'] = key.substr(5).toLowerCase()+"s";
+                    this.ConvertObject(key, object, element, object[key]['@containerField']);
+                } else {
+                    this.ConvertObject(key, object, element);
+                }
             }
         }
         else if ( typeof object[ key ] === "number" )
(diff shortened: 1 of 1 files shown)


coderextreme:
When the JSON is converted to DOM, we need to check to see if we should add a containerField to the DOM.

coderextreme:
This breaks the JSON parser for HAnim, even though it was supposed to support it.  Hmm.

coderextreme:
I must be sleepy, looking at production instead of dev.  I think this patch is needed when converting JSON to XML or DOM, to add containerFields.

andreasplesch:
Let's keep x3dscript separate since it needs extra testing and is not quite ready to be included in a release. Please remove here.

Do we have a small test for HAnim json  ?

andreasplesch:
https://coderextreme.net/X3DJSONLD/src/main/html/joeandjin.html I think is related.

src/util/json/JSONParser.js
```diff
@@ -299,7 +299,27 @@ x3dom.JSONParser.prototype.ConvertToX3DOM = function ( object, parentkey, elemen
...
+                else if ( key.indexOf( "HAnim" ) === 0 && key !== "HAnimHumanoid" && typeof object[ key ][ "@USE" ] != "undefined" )
+
+                {
+                    object[ key ][ "@containerField" ] = key.substr( 5 ).toLowerCase() + "s";
+                    this.ConvertObject( key, object, element, object[ key ][ "@containerField" ] );
+                }
+                else
+                {
+                    this.ConvertObject( key, object, element );
+                }
```
andreasplesch, line 322: cherry picked in #1142

coderextreme:
I think you may be missing the first line?  Or did I mess up?

John

andreasplesch:
Here is the complete change:
https://github.com/x3dom/x3dom/pull/1142/files
I think it is complete ?

coderextreme:
Yeah, looks good! Thanks!

John

coderextreme:
A "small" test for Hanim 1 in X3DOM may be found here:

https://coderextreme.net/X3DJSONLD/src/main/html/x3domjoe.html

Don't use the "Processed" files.

A copy of the JSON Joe files (X3D XML also found in X3d resources
archives), may be found here:
  https://coderextreme.net/X3DJSONLD/src/main/data starting with Joe...

andreasplesch:
https://coderextreme.net/X3DJSONLD/src/main/html/x3domjoe.html

does not require the guards for node==null in addChild.

Do you have an example where the guards are required ?

coderextreme:
I don't know if node==null is still a problem, since JSONParser.js was
probably fixed.

I've patched my X3DOM, but I can unpatch it, given some time to get into a
better mood (sorry, these things happen).

You said:

https://coderextreme.net/X3DJSONLD/src/main/html/joeandjin.html I think is
related.

Is it still? I can work on reducing the number of files being referenced.
 Please give me more time for me to get a coffee replacement,

I'll have some tea to see if I feel like working again.

I understand how it's good to throw exceptions on bad files.

Thanks,

John

andreasplesch:
Please feel free to reopen when appropriate.

PR #1144, opened 2020-11-27
Use alphaCutoff only in 'MASK' alphaMode as according to spec

Sharian:
Hi!

I'm rendering gltf meshes with transparent textures or materials and was wondering why they are not rendering until I changed the value of alphaCutoff. According to the documentation the alphaCutoff should only be used in case alphaMode is "MASK"  [1][2], but it is also used in "BLEND".

After removing the last else clause of:
https://github.com/x3dom/x3dom/blob/cdadbc1c01d3c04989c0bb7176032c90eddca466/src/shader/ShaderDynamic.js#L1310-L1324

my objects got rendered correctly. 

[1] https://doc.x3dom.org/developer/x3dom/nodeTypes/PhysicalMaterial.html
[2] https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/AlphaBlendModeTest/README.md

### DIFF
--- a/src/shader/ShaderDynamic.js
+++ b/src/shader/ShaderDynamic.js
@@ -1317,10 +1317,6 @@ x3dom.shader.DynamicShader.prototype.generateFragmentShader = function ( gl, pro
         shader += "if (color.a <= alphaCutoff) discard;\n";
         shader += "color.a = 1.0;\n";
     }
-    else
-    {
-        shader += "if (color.a <= alphaCutoff) discard;\n";
-    }
 
     //Output the gamma encoded result.
     shader += "if(tonemappingOperator == 1.0) {\n";


andreasplesch:
Looking back

https://github.com/x3dom/x3dom/blame/be67e166b13d8b7c0c20ebd4c87f7ed442a1bce8/src/shader/ShaderDynamic.js#L1274

it seems that ALPHAMASK is supposed to be the test.

Does changing

```js
shader += "if (color.a <= alphaCutoff) discard;\n";
```
back to

```js
shader += "if (color.a <= 0.01) discard;\n";
```

work for your gltf ? Rather than deleting the else clause ? This change works for the AlphaBlendModeTest example.

If so, the question becomes why alphaCutoff becomes set at higher value.

andreasplesch:
Another fix may be to change

https://github.com/x3dom/x3dom/blob/master/src/util/glTF/glTF2Loader.js#L473

to

```js
        alphaCutoff       = material.alphaCutoff || ( alphaMode == "MASK" ? 0.5 : 0.01 ),
```

Could you test this please for your gltf ?

This would allow alphaCutoff effectively only for MASK mode in glTF but allow it for all modes for PhysicalMaterial.

Sharian:
Either fix (reintroducing the epsilon / setting alphaCutoff to 0.1) would work for me and I am also able to set the alphaCutoff in the gltf to 0 to completely circumvent this. I created the MR mostly because of the discrepancy between documentation and implementation. 

In case of the "alphaCutoff effectively only for MASK mode in glTF"-version, the documentation should be updated. (Should I?) 

Is having two different behaviours worth the gained functionality in this case / Do we aim for parity of x3d and gltf to some extend?

andreasplesch:
Thanks, all good points.

Presumably, the overall 0.01 alpha epsilon was originally there for a reason, perhaps just performance ? @tsturm ? So we should probably keep a version of it. Should we make it lower, say 0.005 ?

If so, why not make the epsilon configurable. alphaCutoff is the equivalent. Using alphaCutoff, it would be configurable for PhysicalMaterial. Actually, regular Material has alphaClipThreshold which is used as alphaCutoff.

You are right, this would mean updating the documentation a bit. And it would mean a slight divergence from glTF. But I do not see any real downside. gltf2loader.js can still make sure alphaCutoff is ignored if mode not MASK, for glTF.

Something like

```js
        alphaCutoff       = material.alphaCutoff || 0.5,
        alphaCutoff       = material.alphaMode == 'MASK' ? alphaCutoff : 0.01,
```

I think as long as glTF is loaded per spec., PhysicalMaterial can diverge a bit.

https://www.web3d.org/specifications/X3Dv4Draft/ISO-IEC19775-1v4-WD2/Part01/components/shape.html#PhysicalMaterial

does not have alphaCutoff, so no future conflict there.

Hm, alphaCutoff is set to alphaClipThreshold for regular Material, and for it the default is 0.1 which seems aggressive. Maybe we should change the default to 0.01 there as well.

https://github.com/x3dom/x3dom/blob/cdadbc1c01d3c04989c0bb7176032c90eddca466/src/nodes/Shape/Appearance.js#L153

How does that sound ?

If ok, do you want to make these changes to the PR ? I could do that as well.

- PhysicalMaterial doc: restriction to MASK only for glTF
- gltf2Loader.js: see above
- Appearance.js : potentially change default to 0.01, update doc

andreasplesch:
I thought it over. Let's keep it simple and just remove the general alpha check as in your PR at least until @tsturm wants the 0.01 epsilon back. It seems arbitrary to have that epsilon if it is not configurable. Hopefully, there is not too much of performance impact since without the check more transparent pixels need to be rendered.

This way no other changes are needed.

@tsturm : Should we keep the overall discarding of low alpha pixels ?

Issue #1226, opened 2022-07-24
Removing a <movietexture> dynamically causes errors.


microaaron:
```Removing a <movietexture> dynamically causes errors.```

Uncaught TypeError: Failed to execute 'texImage2D' on 'WebGL2RenderingContext': Overload resolution failed.
    at updateMovie (x3dom.debug.js:21188:16)
https://github.com/x3dom/x3dom/blob/23bb1572587a1a9117280d1da26ffc33bc2f1f2e/src/Texture.js#L449

andreasplesch:
I think this happens because the `<video>` DOM element is removed without emitting an `ended` event. .pause() in shutdown should cause an `ended` event. `ended` may only be emitted after a video started to play. If there is no user interaction it does not start to play.

andreasplesch:
It turns out that shutdown actually should not cause an `ended` playing event because it should not try to loop the video. It seems easiest to just remove the updating in shutdown. I will also add better IOS compatibility, and better auto-starting.

microaaron:
Thanks.

microaaron:
This error still exists. 
https://github.com/x3dom/x3dom/blob/23bb1572587a1a9117280d1da26ffc33bc2f1f2e/src/Texture.js#L449
tex._video=null

andreasplesch:
I could reproduce on Linux, Chrome 103.0.5060.53 . I had only tested on Windows, latest Chrome where it worked. What platform did you use ?

microaaron:
> I could reproduce on Linux, Chrome 103.0.5060.53 . I had only tested on Windows, latest Chrome where it worked. What platform did you use ?

windows，chrome 102.0.5005.115

microaaron:
setInterval() is called twice.
https://github.com/x3dom/x3dom/blob/835ef694bd3c150d7d819029321525f928dc6da9/src/Texture.js#L465-L479
##1228

microaaron:
And...is it necessary for the second parameter of setInterval() to match the frame rate of the video?

andreasplesch:
Thx, a copy and paste mistake.

---on the phone---

microaaron:
Firefox reported two warnings:
[WEBGL_debug_renderer_info is deprecated in Firefox and will be removed. Please use RENDERER. ](https://github.com/x3dom/x3dom/blob/master/src/gfx_webgl.js#L161)
WebGL warning: readPixels: Out-of-bounds reads with readPixels are deprecated, and may be slow.

[...]

PR #1228, opened 2022-07-28
Texture.js: Fix the problem of setInterval() being called repeatedly.

microaaron:
Fix the problem of setInterval() being called repeatedly. #1226


andreasplesch:
Thanks. Let's deal with RigidBodyPhysic separately.

microaaron:
There is to many invalid commits here, let me pull a new request. #1229

PR #1229, opened 2022-07-28
Texture.js: Add a guard to prevent setInterval() from being called repeatedly

microaaron:
Remove invalid commits of #1228 .

### DIFF
--- a/src/Texture.js
+++ b/src/Texture.js
@@ -431,9 +431,6 @@ x3dom.Texture.prototype.updateTexture = function ()
                 tex._video.setAttribute( "autoplay", "" );
                 tex._video.setAttribute( "playsinline", "" );
                 tex._video.crossOrigin = "anonymous";
-                tex._video.retryInterval = 1000;
-                tex._video.retryGrowth = 0.2;
-                tex._video.retryTimeoutID = null;
                 // p.appendChild( tex._video );
                 // tex._video.style.visibility = "hidden";
                 // tex._video.style.display = "none";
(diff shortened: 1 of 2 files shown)


src/nodes/Texturing/Texture.js
```diff
@@ -39,7 +39,7 @@ x3dom.registerNodeType(
             this.addField_SFBool( ctx, "hideChildren", true );
 
             this._video = null;
-            this._intervalID = 0;
+            this._intervalID = null;
```
andreasplesch, line 42: The id is actually a numerical value.

https://developer.mozilla.org/en-US/docs/Web/API/setInterval#return_value

src/Texture.js
```diff
@@ -467,6 +467,11 @@ x3dom.Texture.prototype.updateTexture = function ()
             tex._video.play()
                 .then( function fulfilled ()
                 {
+                    if ( tex._intervalID )
```
andreasplesch, line 470: When does that happen ? Probably startVideo() should not be called in the first place when already updating.

src/nodes/Texturing/Texture.js
microaaron: Yes, setInterval() returns a non-zero value.

src/Texture.js
microaaron: If _intervalID is a non-zero number, it means that there is already a running interval, and it must be closed before setting a new one.

andreasplesch: Yes. I feel the guard may not needed and would rather hide any underlying problem. Could you add a logWarning message ?

src/nodes/Texturing/Texture.js
andreasplesch: I mean initialising it to 0 seems more appropriate rather than null.

microaaron: In fact, there is no difference between 0 and null, I prefer null, because it clearly tells me that there is no interval, and 0 may cause misunderstanding, making people mistakenly think that interval 0 is an existing interval.

src/Texture.js
microaaron: If this guard is removed, it will be difficult to troubleshoot errors caused by repeated calls to startVideo().

microaaron:
@andreasplesch 
[Fixed a bug caused by removing a movietexture during retry](https://github.com/x3dom/x3dom/pull/1229/commits/55d2b889e1857a22813e48799f685a4ac5b1c9d7)

microaaron:
I think it's better to replace timeOut with mouse and keyboard events

src/nodes/Texturing/Texture.js
andreasplesch: ok but note 0 === null
false

andreasplesch:
Thanks !

PR #1235, opened 2022-10-30
Update WebVR code to WebXR

msub2:
Opening a draft PR while I work this out still. Right now I've gotten it to a point where it's successfully spinning up an XRSession and rendering to the headset, but for some reason it's only appearing in the right eye. It's not immediately clear why this is but I'll continue to work on it.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,6 @@
 {
     "parserOptions": {
-        "ecmaVersion": 5
+        "ecmaVersion": 6
     },
     "rules": {
         "indent": ["error", 4, { "SwitchCase": 1 }],
--- a/src/Runtime.js
+++ b/src/Runtime.js
@@ -1472,30 +1472,21 @@ x3dom.Runtime.prototype.onAnimationFinished = function ()
 
 x3dom.Runtime.prototype.enterVR = function ()
 {
-    if ( this.canvas.vrDisplay && !this.canvas.vrDisplay.isPresenting )
-    {
-        this.canvas.vrDisplay.requestPresent( [ { source: this.canvas.canvas } ] ).then( function ()
-        {
-            this.canvas.doc.needRender = true;
-        }.bind( this ) );
-    }
+    this.canvas.enterVR();
 };
 
 x3dom.Runtime.prototype.exitVR = function ()
 {
-    if ( this.canvas.vrDisplay && this.canvas.vrDisplay.isPresenting )
-    {
-        this.canvas.vrDisplay.exitPresent();
-    }
+    this.canvas.exitVR();
 };
 
 x3dom.Runtime.prototype.toggleVR = function ()
 {
-    if ( this.canvas.vrDisplay && !this.canvas.vrDisplay.isPresenting )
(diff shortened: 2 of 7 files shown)


msub2:
After more tinkering I *believe* the rendering issue stems from somewhere in the code for the shaders, somewhere around https://github.com/x3dom/x3dom/blob/master/src/shader/ShaderComposed.js#L75. There are also the various `isVR` codepaths in gfx_webgl.js, I'm just having a bit of difficulty nailing down exactly what's making it only render in the right eye still.

tsturm:
x3dom uses a Instanced-based single pass technique to render both eyes in one frame. I think the problem is that you set the wrong Viewport at https://github.com/msub2/x3dom/blob/webxr/src/Viewarea.js#L175

If you remove this line it should work

msub2:
If I remove that, it looks closer to correct on a flatscreen view (i.e. with a [WebXR API Emulator](https://github.com/felixtrz/WebXRQuestEmulator)) but does not render correctly in-headset (my test scene contains a cube positioned directly ahead of the camera, but currently it renders distorted and in the far bottom-left of my view).

Flatscreen view


In-headset view (this also requires looking down and to the left in real life)

andreasplesch:
great progress !

msub2:
Appreciate the encouragement! I'm always juggling several different projects at once but I'll try and continue to chip away at this as I can.

andreasplesch:
Would it make sense to try the 'local' reference space ?

```js
 session.requestReferenceSpace("local").then(function ( space )
```

tsturm:
I already have a version which works on my old Oculus Go, I will push it to a webxr branch as soon as possible.

tsturm:
I have pushed my current state to a new webxr branch for testing:

https://github.com/x3dom/x3dom/tree/webxr

msub2:
@tsturm Fantastic, will take a look at it later today.

msub2:
Rendering does indeed work on that branch, reset mine to it and fixed a small issue that was causing a DOMException. Controller input looks to be the main thing that still needs conversion so I'll look at that next, though I do have a couple questions/thoughts:

- @tsturm Is it intended for the VR Button to not have any content (hence making it unclickable) by default? I'm specifically referencing this function here: https://github.com/x3dom/x3dom/blob/163bf1d672304632795efc59f0fb02b05472d4e5/src/X3DCanvas.js#L1201
- I notice that currently the library is using assets hosted on the x3dom website for controller models. You could take advantage of the [WebXR Input Profiles repo](https://github.com/immersive-web/webxr-input-profiles), specifically the assets package, and have a complete collection of official controller models to be utilized by the library.

rghv96:
Hello @tsturm @msub2 @andreasplesch, I tried to test this pull request and found issues which I have highlighted here https://github.com/x3dom/x3dom/issues/1245.  Please let me know if I am missing something to resolve the issues.

msub2:
Closing now that https://github.com/x3dom/x3dom/commit/18cf88c7d3fec2719846d055a4433018396a4b5a has been merged

PR #1257, opened 2023-02-24
Add viewpoint and navigation support in webxr

rghv96:


rghv96:
@andreasplesch @tsturm @msub2

Please review this PR.

andreasplesch:
Could you please start from the current webxr branch, so that only your changes are listed in this or a new PR ?

rghv96:
@andreasplesch I have added the scaleFactor of 0.001 in the fit function in VRControllerManager. Also, I don't fully understand why do we need this function as ideally, we should not need any movement in viewArea when we enter VR mode.

andreasplesch:
I think it was an attempt to automatically have a better view when switching to XR. It may not be needed with viewpoint support. I used the small scale factor only as a safe way to disable the fit function without removing it.

rghv96:
Alright, I think the PR is ready to review @andreasplesch

andreasplesch:
@tsturm and @msub2 ?

msub2:
Apologies for delay, looks good to me though

andreasplesch:
Thanks.

PR #1287, opened 2023-08-05
Turntable init

andreasplesch:


andreasplesch:
fix feedback at init for turntable with animateTo

PR #1312, opened 2023-10-15
fog w shadows & ssao

dmorehead:
simplified "fog + shadows" to no longer have to use a custom GL_BLEND or use gl.drawArrays twice within the shadow pass. Instead use fog to "mask" the shadows so only the "shadows - fog" are returned from renderShadows. Also applied this same logic to the SSAO pass as it feels more natural visually to have the SSAO disappear into the fog.
fogShadSSAO


andreasplesch:
Thanks ! I think this could work well as simplifications are always welcome. Please run `npm run lint-fix` which should take care of the formatting. There may also be some comments on the code.

src/SSAO.js
```diff
@@ -255,6 +255,12 @@ x3dom.SSAO.blur = function ( stateManager, gl, scene, ssaoTexture, depthTexture,
 
     stateManager.useProgram( sp );
 
+    // Mask ssao w/ fog, no need for fogColor
+    // ALWAYS pass FOG uniforms into shader or scenes w/out fog will barf
```
andreasplesch, line 259: Could you check for fog here and in the SSAO shader (properties.FOG, see below) ?

src/shader/ShaderDynamic.js
```diff
@@ -742,19 +742,33 @@ x3dom.shader.DynamicShader.prototype.generateFragmentShader = function ( gl, pro
...
+    if ( properties.LIGHTS || properties.FOG || properties.CLIPPLANES )
     {
+        shader += "uniform vec3 eyePosition;\n";
         shader += "varying vec4 fragPosition;\n";
-        shader += "varying vec4 fragPositionWS;\n";
         shader += "uniform float isOrthoView;\n";
-    }
+        shader += "varying vec4 fragPositionWS;\n";
 
+        if ( properties.FOG )
```
andreasplesch, line 766: seems cleaner to keep outside and not to nest inside here.

src/shader/ShaderDynamic.js
```diff
@@ -742,19 +742,33 @@ x3dom.shader.DynamicShader.prototype.generateFragmentShader = function ( gl, pro
...
+    // {
+    //     shader += "varying vec4 fragPosition;\n";
+    //     shader += "varying vec4 fragPositionWS;\n";
+    //     shader += "uniform float isOrthoView;\n";
+    // }
+
+    // same as vertex shader but with fragPositionWS for fogNoise (w/ or w/out lights)
+    if ( properties.LIGHTS || properties.FOG || properties.CLIPPLANES )
     {
+        shader += "uniform vec3 eyePosition;\n";
```
andreasplesch, line 761: Is this needed ?

src/SSAO.js
dmorehead: ideally yes, however (unlike shapes & shadows) properties are not being created in the ssao chain, and I was hesitant to shoehorn them into x3dom.Utils.wrapProgram.

src/shader/ShaderDynamic.js
dmorehead: originally I nested this check to mimic the exact logic used to create the corresponding vertex shader, but looks fine to moved it back outside.

src/shader/ShaderDynamic.js
dmorehead: looks like it is only needed by the vertex shader, everything seems to work w/out it.

andreasplesch (approved):

src/shader/ShaderShadowRendering.js
```diff
@@ -172,15 +163,25 @@ x3dom.shader.ShadowRenderingShader.prototype.generateFragmentShader = function (
...
     // gamma coefficient), i.e. the umbra is corrected for now, the penumbra
     // is incorrect and full light is zero here so unaffected as well.
+    shader += "    color = " + x3dom.shader.encodeGamma( {}, "vec4(shadowValue, shadowValue, shadowValue, 1.0)" ) + " ;\n";
+
+    if ( properties.FOG )
+    {
+    // mask shadow w/ fogDepth, visually no need for fogColor
+        shader += "   vec3 fragEyePosition = eyeCoords.xyz / eyeCoords.w;\n" +
+              "   float f0 = calcFog(fragEyePosition);\n" +
+        //"   color.rgb = fogColor * (1.0 - f0) + f0 * color.rgb;\n";
```
andreasplesch, line 178: Please remove comment.

andreasplesch (changes requested):
Please add a line to CHANGELOG.md

src/shader/ShaderShadowRendering.js
andreasplesch: Sorry, I meant the comment in line 178.

andreasplesch (approved):

PR #1313, opened 2023-10-16
Fog plus height and noise

dmorehead:
Optional fogHeight and fogNoise. Not setting them will maintain the original fog behavior. The height is best seen when lowering the visibility range. here is a working pen. https://codepen.io/mrmernan/pen/KKbYrdJ


andreasplesch:
Please resolve conflicts. Probably easiest to `git pull` or sync master from github, and use a new branch.

andreasplesch:
Do you think it is possible to directly use fogHeight to normalize the WS .y component, rather than fogRange ? It would be good to decouple vertical fog gradients from view direction fog gradients.

dmorehead:
good suggestion, normalized fogHeight separately from fogRange. Its a bit less code and the result seems more visually predictable. Also updated the pen with these changes (codepen has a 1MB limit so I've only included x3dom.js, not x3dom-full.js)

src/nodes/EnvironmentalEffects/X3DFogNode.js
```diff
@@ -67,6 +67,30 @@ x3dom.registerNodeType(
              * @instance
              */
             this.addField_SFFloat( ctx, "visibilityRange", 0 );
+
+            /**
+             * FogHeight is in world scale like visibilityRange.
```
andreasplesch, line 72: `fogHeight is the maximum height of fog above 0 in world scale. A value of zero disables the effect.` Is that accurate ?

src/nodes/EnvironmentalEffects/X3DFogNode.js
```diff
@@ -67,6 +67,30 @@ x3dom.registerNodeType(
...
+             * @var {x3dom.fields.SFFloat} fogHeight
+             * @memberof x3dom.nodeTypes.X3DFogNode
+             * @initvalue 0
+             * @range [0, -inf]
```
andreasplesch, line 76: `[0, inf]`

src/nodes/EnvironmentalEffects/X3DFogNode.js
```diff
@@ -67,6 +67,30 @@ x3dom.registerNodeType(
...
+             * @field x3dom
+             * @instance
+             * Hint: Start with a low visibilityRange (very dense fog) to adjust fogHeight.
+             */
+            this.addField_SFFloat( ctx, "fogHeight", 1000000.0 );
```
andreasplesch, line 81: Should the default value be `0` ?

andreasplesch:
> good suggestion, normalized fogHeight separately from fogRange. Its a bit less code and the result seems more visually predictable. Also updated the pen with these changes (codepen has a 1MB limit so I've only included x3dom.js, not x3dom-full.js)

The pen looks good. Did you want to incorporate these changes in the PR as well, eg. commit them to the branch ?

src/nodes/EnvironmentalEffects/X3DFogNode.js
andreasplesch: I think it is correct.

src/shader/ShaderDynamic.js
```diff
@@ -1350,11 +1347,24 @@ x3dom.shader.DynamicShader.prototype.generateFragmentShader = function ( gl, pro
...
-        shader += "color.rgb = fogColor * (1.0-f0) + f0 * (color.rgb);\n";
+        shader +=   "  float f0 = calcFog(fragEyePosition);\n" +
+                    //"  color.rgb = fogColor * (1.0-f0) + f0 * (color.rgb);\n";
+                    // original fog above, or optional Noise and Height below
+                    "  if (fogNoise > 0.0) f0 += mix( 0.0, 0.1, noise3d(fragPositionWS.xyz * fogNoise) );\n" +
+                    "    vec3 fcol = fogColor * (1.0-f0) + f0 * (color.rgb);\n" +
+                    "  if (fogHeight == 0.0) {\n" +
+                    "    color.rgb = mix( fcol, color.rgb, f0 );\n" +
+                    "  } else if (fogHeight > 0.0) {\n" +
+                    "    float f1 = clamp( 1.0 - ((abs(fogHeight)-fragPositionWS.y) / abs(fogHeight)), 0.0, 1.0);\n" +
```
andreasplesch, line 1360: `abs(fogHeight)` here always `fogHeight` since positive. So use just `fogHeight`

src/shader/ShaderDynamic.js
```diff
@@ -1350,11 +1347,24 @@ x3dom.shader.DynamicShader.prototype.generateFragmentShader = function ( gl, pro
...
+                    "    color.rgb = mix( fcol, color.rgb, max(f1,f0) );\n" +
+                    "  } else if (fogHeight < 0.0) {\n" +
+                    "    float f1 = clamp(       ((abs(fogHeight)-fragPositionWS.y) / abs(fogHeight)), 0.0, 1.0);\n" +
```
andreasplesch, line 1363: Since fogHeight negative, this is:
`(-fogHeight-fragPositionWS.y)/-fogHeight =
(fogHeight+fragPositionWS,y)/fogHeight`
simplified. 
May need more thinking.

src/shader/ShaderDynamic.js
```diff
@@ -1350,11 +1347,24 @@ x3dom.shader.DynamicShader.prototype.generateFragmentShader = function ( gl, pro
...
+                    // de-couple fogHeight from fogRange
+                    //"  color.rgb = mix( fcol, color.rgb, max(f1,f0) );\n" +
+                    "    color.rgb = mix( fcol, color.rgb, min(f1,f0*2.0) );\n" +
```
andreasplesch, line 1366: f0*2.0  ? May need clear thinking.

src/shader/ShaderDynamic.js
```diff
@@ -1350,11 +1347,24 @@ x3dom.shader.DynamicShader.prototype.generateFragmentShader = function ( gl, pro
...
+                    //"  color.rgb = fogColor * (1.0-f0) + f0 * (color.rgb);\n";
+                    // original fog above, or optional Noise and Height below
+                    "  if (fogNoise > 0.0) f0 += mix( 0.0, 0.1, noise3d(fragPositionWS.xyz * fogNoise) );\n" +
+                    "    vec3 fcol = fogColor * (1.0-f0) + f0 * (color.rgb);\n" +
+                    "  if (fogHeight == 0.0) {\n" +
+                    "    color.rgb = mix( fcol, color.rgb, f0 );\n" +
+                    "  } else if (fogHeight > 0.0) {\n" +
+                    "    float f1 = clamp( 1.0 - ((abs(fogHeight)-fragPositionWS.y) / abs(fogHeight)), 0.0, 1.0);\n" +
+                    "    color.rgb = mix( fcol, color.rgb, max(f1,f0) );\n" +
+                    "  } else if (fogHeight < 0.0) {\n" +
```
andreasplesch, line 1362: just `else`
The remaining `if (..)` is not necessary. In the shader code branching logic should be avoided as much as possible since it is considered slow.

PR #1345, opened 2025-09-10
Flip the Texture generated by PixelTexture

microaaron:
Simplify the code.
No change in functionality.
More compliant with the specification.(Ref. https://registry.khronos.org/webgl/specs/latest/1.0/#texImage2D and https://www.web3d.org/documents/specifications/19775-1/V4.0/Part01/fieldTypes.html#SFImageAndMFImage)

[WebGL Specification](https://registry.khronos.org/webgl/specs/latest/1.0/#5.14.8): The first pixel transferred from the source to the WebGL implementation corresponds to the upper left corner of the source. This behavior is modified by the UNPACK_FLIP_Y_WEBGL [pixel storage parameter](https://registry.khronos.org/webgl/specs/latest/1.0/#PIXEL_STORAGE_PARAMETERS), except for ImageBitmap arguments, as described in the abovementioned section.
[X3DOM Specification](https://www.web3d.org/documents/specifications/19775-1/V4.0/Part01/fieldTypes.html#SFImageAndMFImage): Pixels are specified from left to right, bottom to top. The first hexadecimal value is the lower left pixel and the last value is the upper right pixel.


andreasplesch:
Sorry for the delayed response.
This seems like a good, simplifying idea. Let's find tests for the diffusemapchannel case.

andreasplesch:
https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Assets/c4f44afba6af5c9246e1b81b3178ee8efe833f2d/Models/TextureTransformMultiTest/glTF/TextureTransformMultiTest.gltf

seems like a good test as mentioned in #1350. It works fine with #1350 applied together with this.

Ok, let's go for it. Could you please add an entry to CHANGELOG.md ? Thanks.

PR #1346, opened 2025-09-11
Set the acceptable type of SFNode and MFNode to an array

microaaron:
Some fields can accept multiple types of nodes, such as X3DPickSensorNode.pickTarget, HAnimHumanoid.skeleton.
So set the acceptable type of SFNode and MFNode to an array.
487961718-cc83dbe7-655a-4ae9-8dd8-40982ac74bc1
487961719-4a475203-3814-4eac-87ed-a0dd408e6174
487961720-3e711c60-b968-45d6-805e-6fccdff21b00

No change in functionality.
More compliant with the specification.


andreasplesch:
Sorry for the long delay. This would be a deeper change so likely needs thorough testing. It will probably work fine. Do you have a proto where this would be required ?
Also, there may be places (HAnim?, Picking?)  where `x3dom.isa` would benefit from/need to changing to `x3dom.isOneOf` ?

coderextreme:
I know in JSON, SFNode becomes object and MFNode becomes array. But X3DOM has no JSON output and on input JSON is converted to DOM to match X3DOM.

John

